PR #460, opened 2020-02-12
Convert misc cost fields to alphanumeric; add "Deadline to consign" field

pepopowitz:

### New field added

### Existing fields updated

```
(5 rows)
```

```
(5 rows)
```

image

image


pepopowitz:
Closing because GitHub got kind of wonky when we marked artsy/convection as public, and couldn't find my changes. I opened https://github.com/artsy/convection/pull/465 instead.

PR #465, opened 2020-02-13
Modify offer fields to better accommodate user needs

pepopowitz:

### New field added

### Existing fields updated

```
(5 rows)
```

```
(5 rows)
```

image

image


db/migrate/20200211225056_modify_offers_table.rb
```diff
@@ -0,0 +1,87 @@
...
+      t.column(:insurance_percent, :float)
+      t.column(:other_fees_percent, :float)
+
+      t.remove(:deadline_to_consign)
+
+      t.change(
+        :photography_info,
+        :integer,
+        using:
+          "CASE WHEN (photography_info ~ '^\d+$') THEN photography_info::integer ELSE null END"
```
pepopowitz: While this _could_ preserve data when reverting the migration, it wouldn't actually match any of the data that was created in the `up`. These are mostly here because if we don't do _something_, it'll try to convert all the just-plain-text values to integer and fail.

spec/controllers/admin/offers_controller_spec.rb
```diff
@@ -297,16 +294,14 @@
...
           )
         put :update,
             params: {
               id: retail_offer.id,
-              offer: { insurance_dollars: nil, insurance_percent_whole: nil }
+              offer: { insurance_info: nil }
             }
-        expect(retail_offer.reload.insurance_cents).to eq nil
-        expect(retail_offer.reload.insurance_percent_whole).to eq nil
+        expect(retail_offer.reload.insurance_info).to eq ""
```
pepopowitz: I wasn't expecting to have to change the right side of this assertion (from `.to eq nil` to `.to eq ""`). Maybe a reviewer can tell me why I would have had to do this?

app/helpers/offers_helper.rb
```diff
@@ -103,50 +103,6 @@ def commission_display(offer)
...
-
-    if offer.other_fees_cents.present?
-      "#{offer.currency} #{
-        Money.new(offer.other_fees_cents, offer.currency).format
-      }"
-    else
-      "#{offer.other_fees_percent * 100}%"
-    end
-  end
-
```
damassi: ✌️

pepopowitz:
Closing this in favor of #466 because I learned that we need to submit PRs from within artsy/convection, not our personal forks.

db/migrate/20200211225056_modify_offers_table.rb
```diff
@@ -0,0 +1,87 @@
...
+            "THEN TRIM(TO_CHAR(insurance_percent * 100, '99D00%')) END"
+      )
+      t.change(
+        :other_fees_info,
+        :string,
+        using:
+          'CASE WHEN (other_fees_info IS NOT NULL) ' \
+            "THEN format('%s %s', TRIM(to_char(other_fees_info / 100, '99999')), currency) " \
+            'WHEN (other_fees_percent IS NOT NULL) ' \
+            "THEN TRIM(TO_CHAR(other_fees_percent * 100, '99D00%')) END"
```
damassi: 👍 Curious if doing this in SQL is indeed the ideal way? Or if a view helper would be a better place? I guess in this case it does make all the data consistent, but also a little less flexible. (That said, I know very little SQL.)

PR #466, opened 2020-02-13
Modify offer fields to better accommodate user needs

pepopowitz:

### New field added

### Existing fields updated

```
(5 rows)
```

```
(5 rows)
```

image

image


db/migrate/20200211225056_modify_offers_table.rb
```diff
@@ -0,0 +1,87 @@
...
+      t.column(:insurance_percent, :float)
+      t.column(:other_fees_percent, :float)
+
+      t.remove(:deadline_to_consign)
+
+      t.change(
+        :photography_info,
+        :integer,
+        using:
+          "CASE WHEN (photography_info ~ '^\d+$') THEN photography_info::integer ELSE null END"
```
pepopowitz: While this _could_ preserve data when reverting the migration, it wouldn't actually match any of the data that was created in the `up`. These are mostly here because if we don't do _something_, it'll try to convert all the just-plain-text values to integer and fail.

spec/controllers/admin/offers_controller_spec.rb
```diff
@@ -297,16 +294,14 @@
...
           )
         put :update,
             params: {
               id: retail_offer.id,
-              offer: { insurance_dollars: nil, insurance_percent_whole: nil }
+              offer: { insurance_info: nil }
             }
-        expect(retail_offer.reload.insurance_cents).to eq nil
-        expect(retail_offer.reload.insurance_percent_whole).to eq nil
+        expect(retail_offer.reload.insurance_info).to eq ""
```
pepopowitz: I wasn't expecting to have to change the right side of this assertion (from `.to eq nil` to `.to eq ""`). Maybe a reviewer can tell me why I would have had to do this?

db/migrate/20200211225056_modify_offers_table.rb
```diff
@@ -0,0 +1,87 @@
...
+      t.rename(:shipping_cents, :shipping_info)
+      t.rename(:insurance_cents, :insurance_info)
+      t.rename(:other_fees_cents, :other_fees_info)
+
+      t.change(
+        :photography_info,
+        :string,
+        using:
+          'CASE WHEN (photography_info IS NOT NULL) ' \
+            "THEN format('%s %s', TRIM(to_char(photography_info / 100, '99999')), currency) END"
```
damassi: 👍 Curious if doing this in SQL is indeed the ideal way, or if a view helper would be a better place? I guess in this case it does make all the data consistent, but also (seems) a little less flexible. (That said, I know very little in terms of idioms for data transforms like this.)

db/migrate/20200211225056_modify_offers_table.rb
jonallured: FWIW, I feel that you've gone above and beyond here - I would have done something like this:

```ruby
raise ActiveRecord::IrreversibleMigration, "Can't go back!"
```

Your way is certainly better. I guess I'm just noting that it's a valid strategy to consider throwing up one's hands on the `down` method. 😝 

But again bravo for being extremely thorough!! 🤗 💯

spec/controllers/admin/offers_controller_spec.rb
jonallured: My guess is that this has to do with how Rails parses string vs integer params? We could dig into it if it becomes an issue but it's probably fine. 🤷‍♂

db/migrate/20200211225056_modify_offers_table.rb
jonallured: Yeah I mean what I expected to see what Ruby, but damn if you know how to do this in SQL then I'm super impressed!! 🏆 

The only lingering thought I have is how we can be sure this works correctly outside of running it on staging and doing some QA on the results. I mean I would write a test for this in the abstract but there really aren't good options for that type of thing when it comes to Rails migrations. I think the pragmatic thing to do is just try it on staging and then be sure we have DB backups went it comes time to promote to production in case something wacky happens.

Thoughts?

db/migrate/20200211225056_modify_offers_table.rb
pepopowitz: I thought I _was_ throwing up my hands on the `down` method 😅

spec/controllers/admin/offers_controller_spec.rb
pepopowitz: It's no concern for me if it's no concern for you. 👍

db/migrate/20200211225056_modify_offers_table.rb
pepopowitz: @damassi Re: flexibility, I'm not sure what you mean. It's a one-time conversion, so I wouldn't anticipate needing any flexibility to make changes. I suspect you might mean something different though.

pepopowitz: @jonallured Re: expecting Ruby - that makes sense; I have a pretty deep (but secret 🤫) love for SQL, so it's the tool I'd reach for in this kind of situation. This is also the approach I most quickly discovered when researching how to make changes in a migration that are based on existing data. I'm happy to revisit this if you think raw SQL is a bad idea.

jonallured: I do not think this is a bad idea. In fact doing it in Ruby has some tricky edges too. Like what if we rename these things in a future version of the code base? Your SQL version will always work because it's based on the schema while code based off an assumption of the Modeling could be more fragile. Tradeoffs, amiright?

pepopowitz: And Re: verification - I took a look at the data that's in staging & prod when I wrote this migration. There isn't _a lot_ of data in these fields, but there is _some_ data, which is why I felt it important to make sure we preserve it. 

I verified that the queries I was running inside the `using:` clause returned the expected value for all current data. I also tried to simulate all the possible staging/prod data in my local environment (see "Data snapshot before the migration" and "Data snapshot after the migration" in the PR body), and ran the migration probably a dozen or so times before feeling confident that it was converting things properly.

I definitely agree we should keep an eye on what happens in staging, but I also feel pretty confident that my manual tests covered things.

pepopowitz:
Looks like I have to do some rubocop fixes too, but in the meantime, I ran the migration locally against a copy of the staging database (thanks @jonallured and #470!). It succeeded! Some data:

These queries are against prod, just to verify that the counts are in line with what's in staging: 

```
convection=> select count(1) from offers where shipping_cents is not null;
 count 
-------
     0
(1 row)

convection=> select count(1) from offers where photography_cents is not null;
 count 
-------
    39
(1 row)

convection=> select count(1) from offers where insurance_cents is not null;
 count 
-------
     2
(1 row)

convection=> select count(1) from offers where other_fees_cents is not null;
 count 
-------
     7
(1 row)

convection=> select count(1) from offers where other_fees_percent is not null;
 count 
-------
     5
(1 row)

convection=> select count(1) from offers where insurance_percent is not null;
 count 
-------
   204
(1 row)
```

And these are the contents in my local db after converting my copy of the staging db:

```
convection_development=# select distinct(photography_info), count(1) from offers group by photography_info;
 photography_info | count 
------------------+-------
 0 USD            |     1
 200 USD          |     3
 100 USD          |    20
 150 GBP          |     1
 5 GBP            |    13
 10 GBP           |     1
                  |  3355

convection_development=# select distinct(shipping_info), count(1) from offers group by shipping_info;
 shipping_info | count 
---------------+-------
               |  3394
(1 row)

convection_development=# select distinct(insurance_info), count(1) from offers group by insurance_info;
 insurance_info | count 
----------------+-------
 1 USD          |     1
 1.75%          |    12
 2.00%          |     3
                |  3189
 .50%           |     1
 0 GBP          |     1
 20.00%         |     1
 1.50%          |   186
(8 rows)

convection_development=# select distinct(other_fees_info), count(1) from offers group by other_fees_info;
 other_fees_info | count 
-----------------+-------
 1 USD           |     4
 0 USD           |     1
                 |  3383
 0 EUR           |     1
 50 USD          |     1
 1.50%           |     3
 10.00%          |     1
(7 rows)
```

sweir27:
Cool! Glad these fields will be more often used in their new form.

I'm a little 😢because with this change we remove any potential ability to calculate the total amount of an offer (but sounds like we don't often _have_ the data in that form). So, I understand this change and the current use-case, but I hope someday our user needs allow us to structure it better! Did you all consider keeping the existing and simply _adding_ the free text fields?

One thought on `deadline_to_consign`-- could that be a proper date field? Or would we consider having a date field _and_ a `deadline_to_consign_note` field? Just thinking if we _can_ get a proper date there, we'll be able to eventually do cool things like send reminders. It's sort of frustrating to migrate from a text field -> a structured field after-the-fact.

(It still brings me SO MUCH joy to see you all diving into convection 💪!!!)

pepopowitz:
@sweir27 the data type changes are largely driven by the way the app is being used. The [queries in my last comment](https://github.com/artsy/convection/pull/466#issuecomment-586385271) show how rare it is that we actually have a numeric value for any of these fields. I'll raise the concern of not being able to calculate a total to the main stakeholder (Agnieszka), but my understanding is that she can't really use the fields as-is, and is mostly just putting everything in the notes field as a workaround. My take is that if we decide in the future that we want more structured fields, we can add them back, but right now they aren't useful.

I'll raise the concerns to Agnieszka and hold off on merging until I get confo from her and @jonallured that this is okay to move forward with.

sweir27:
@pepopowitz Totally! I get that.

Just hoping to think through if there are any ways (or small wins-- like the deadline date) to achieve _everything_ we want 😅(high-quality, structured data + easy admin flow).

joeyAghion:
I share some of @sweir27's concerns about dropping the _capacity_ for structured data. Even if they're 100% unused, could we _add_ new "notes" or "other" fields that enable flexible comments, while retaining the original fields? (This would obviate the epic SQL migration which to be clear I _respect_ but I also _fear_.)

I'm glossing over how the front-end might need modifying to support both the structured and supplemental fields, but hoping that wouldn't be too complex. E.g., they could be concatenated or coalesced.

The `deadline_...` string column also worries me. We shouldn't actually encourage entries that are relative to the current date, right? Those would start to lose meaning or even mislead over time.

jonallured:
My preference will always be towards avoiding over engineering and I'm concerned that some of this feedback is headed in that direction. This PR feels like a course-correction to me where we're meeting the needs we have today. I don't see a lot of risk here and would prefer to deal with different/new requirements in an agile way if or when they come up. Premature optimization is the root of all evil!

sweir27:
Makes sense! My goal is certainly not to over-engineer, merely to discuss the options. 😄

PR #470, opened 2020-02-14
CSGN-28: Add script to pull staging data down

jonallured:
This PR adds a script to the bin folder that will use `pg_dump` and `pg_restore` to grab staging data and clobber your local dev database with it. It's a blunt instrument at this point, but works great!

Note: you have to be connected to the staging VPN or this will fail.

### DIFF
--- a/bin/pull_data
+++ b/bin/pull_data
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -ex
+
+environment="staging"
+dump_file="$environment.dump"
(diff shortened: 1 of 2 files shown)


jonallured:
@pepopowitz good call on some docs - I ended up starting a `docs` folder and wrote up a basic doc for our custom scripts and this new one in particular. My hope is that as people use these things we'll improve the documentation but at least there's something to start with!

PR #478, opened 2020-02-19
Add scope for available submissions

jonallured:
This PR introduces a concept called "available submissions" where the Submission record is approved, but we have no consignment. This allows us to query convection for these types of Submissions and populate the list/detail pages in Volt.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -70,6 +70,8 @@ class Submission < ApplicationRecord
   scope :completed, -> { where.not(state: 'draft') }
   scope :draft, -> { where(state: 'draft') }
   scope :submitted, -> { where(state: 'submitted') }
+  scope :available,
+        -> { where(state: APPROVED, consigned_partner_submission_id: nil) }
 
   dollarize :minimum_price_cents
 
--- a/spec/models/submission_spec.rb
+++ b/spec/models/submission_spec.rb
@@ -33,6 +33,26 @@ describe Submission do
         expect(Submission.completed.count).to eq(0)
       end
     end
+
+    describe 'available' do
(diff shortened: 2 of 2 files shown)


app/models/submission.rb
```diff
@@ -70,6 +70,8 @@ class Submission < ApplicationRecord
   scope :completed, -> { where.not(state: 'draft') }
   scope :draft, -> { where(state: 'draft') }
   scope :submitted, -> { where(state: 'submitted') }
+  scope :available,
```
damassi: Mind briefly explaining what a `scope` is? (I will also read docs, but i think in your own words TLDR explainations are also super useful for concepts.)

jonallured: Sure! It's a convenient way to specify common queries. It's a shortcut for defining a method - check it out:

```ruby
class Submission < ApplicationRecord::Base
  scope(:available) -> { where(state: APPROVED, consigned_partner_submission_id: nil) }

  # that's a shortcut for doing this...

  def self.available
    where(state: APPROVED, consigned_partner_submission_id: nil)
  end
end
```

Those two are the same. If that Ruby syntax is weird, think of it as two arguments to a method called `scope` - the first is a symbol that is the name of a method we're going to define. The second argument (sometimes called a [stabby lambda](https://www.honeybadger.io/blog/using-lambdas-in-ruby/)) is what's evaluated when the method is called.

This is a common enough pattern that Rails gives us a helper to make these things a bit easier and then it's nice because you can search a file for `scope` and get a sense of the important queries for a given model.

Here's the guides section on it:

https://guides.rubyonrails.org/active_record_querying.html#scopes

PR #498, opened 2020-03-02
Upgrade sidekiq to latest

jonallured:
This PR aims to upgrade convection to the latest sidekiq, here's the doc for that:

https://github.com/mperham/sidekiq/blob/master/6.0-Upgrade.md

I read through it and don't see anything that should trip us up, but I intend to do some QA on staging before promoting this to production.


jonallured:
Ok going to merge, set a deploy block on horizon and then QA this on staging before promoting to production.

PR #504, opened 2020-03-05
[Stitching] Ensure schema matches existing MPv2 definitions

damassi:
**Related**: https://github.com/artsy/metaphysics/pull/2221

This PR updates our GraphQL schema to match the currently _unstitched_ MPv2 schema. We had to make these changes on the convection side because to enable stitching we had to take old mobile app versions into account, and if we modified the schema on the MP side it would be a breaking change. 

There's some `camelCase` versus `snake_case` conversion that we had to do to line things up, but nothing too severe or blocking (there were only three mutations to update).

The other thing we had to change was the GraphQL type names. There were slight differences between what MP uses and what convection had defined. 

Follow-up steps will be to open an additional PR to MP that removes Convection code unrelated to stitching, including three manually defined mutations that communicate with Convection via the REST API.

### DIFF
--- a/app/graph/mutations/update_submission_mutation.rb
+++ b/app/graph/mutations/update_submission_mutation.rb
@@ -1,28 +1,28 @@
 # frozen_string_literal: true
 
 module Mutations
-  module UpdateConsignmentSubmission
+  module UpdateSubmissionMutation
     Definition =
       GraphQL::Relay::Mutation.define do
-        name 'UpdateConsignmentSubmission'
+        name 'UpdateSubmissionMutation'
 
         input_field :id, !types.ID
-        input_field :additional_info, types.String
-        input_field :artist_id, types.String
-        input_field :authenticity_certificate, types.Boolean
+        input_field :additionalInfo, types.String
+        input_field :artistID, types.String
+        input_field :authenticityCertificate, types.Boolean
         input_field :category, Types::CategoryType
         input_field :currency, types.String
         input_field :depth, types.String
-        input_field :dimensions_metric, types.String
+        input_field :dimensionsMetric, types.String
         input_field :edition, types.Boolean
-        input_field :edition_number, types.String
-        input_field :edition_size, types.Int
+        input_field :editionNumber, types.String
+        input_field :editionSize, types.Int
         input_field :height, types.String
-        input_field :location_city, types.String
-        input_field :location_country, types.String
-        input_field :location_state, types.String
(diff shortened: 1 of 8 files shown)


app/graph/mutations/update_submission_mutation.rb
```diff
@@ -1,42 +1,45 @@
...
     Definition =
       GraphQL::Relay::Mutation.define do
-        name 'UpdateConsignmentSubmission'
+        name 'UpdateSubmissionMutation'
 
         input_field :id, !types.ID
-        input_field :additional_info, types.String
-        input_field :artist_id, types.String
-        input_field :authenticity_certificate, types.Boolean
+        input_field :additionalInfo, types.String
```
damassi: `graphql-ruby` has a nice (or rather, confusing) feature where it will automatically take `snake_case` fields and camelize them. However, this doesn't apply to input types.

app/graph/mutations/update_submission_mutation.rb
```diff
@@ -49,9 +52,12 @@ def self.resolve(_obj, args, context)
         raise(GraphQL::ExecutionError, 'Submission Not Found')
       end
 
+      # FIXME: Why does the API reject this property?
+      params.delete('dimensions_metric')
```
damassi: Not sure whats up here, but Rails will reject a submission if this property is included. Looking in Admin UI, this field isn't found, only a naked "dimension: 222x222x222". Perhaps related?

mzikherman:
@pepopowitz if I'm understanding right, the schema here in Convection is changing in a breaking way (to any consumers of the GraphQL schema in Convection).

It's just that, it isn't stitched into Metaphysics (so the iOS app uses Metaphysics schema -> Convection data loaders using the REST endpoints). The web app directly uses the Convection data loaders.

So, making breaking changes to the Convection GraphQL schema is necessary in order to be able to stitch in without breaking changes to the Metaphysics consumers of that schema (just the iOS app).

Implicit in this is the assumption/discovery(?!) that there are _no_ consumers of the current Convection GraphQL API, and so thus this is ok to break.

Might be possible to confirm no consumers of the GraphQL API endpoint in Papertrail or something?

damassi:
> Implicit in this is the assumption/discovery(?!) that there are no consumers of the current Convection GraphQL API, and so thus this is ok to break.

That's the assumption, and if indeed there are consumers those we can change without worrying too much -- but we _cant_ change MP as thats where our mobile client connects to.

PR #505, opened 2020-03-05
[Toolchain] Add guard-rspec

damassi:
Adds some niceties to our testing toolchain by automatically running tests associated with changed files, as well as adding the ability to focus on specific tests: 

1) If a source changes (`foo.rb`), and there's an associated test file (`foo_spec.rb`), run tests.
1) If a test changes, automatically run it
1) If only wanting to run a specific test or test group, add `focus: true` to it 

```rb
it "works", focus: true do 
  ...
end
```

### Use

To start the watcher, run

```rb
$ bundle exec guard
```

onchange

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -6,3 +6,13 @@ guard 'livereload', port: '5003', grace_period: 0.5 do
   watch(%r{app/helpers/.+})
   watch(%r{app/views/.+})
 end
+
+guard :rspec, cmd: 'bundle exec rspec' do
+  watch('spec/spec_helper.rb')                        { "spec" }
+  watch('app/controllers/application_controller.rb')  { "spec/controllers" }
+  watch(%r{^spec/.+_spec\.rb$})
+  watch(%r{^app/(.+)\.rb$})                           { |m| "spec/#{m[1]}_spec.rb" }
+  watch(%r{^app/(.*)(\.erb|\.haml|\.slim)$})          { |m| "spec/#{m[1]}#{m[2]}_spec.rb" }
+  watch(%r{^lib/(.+)\.rb$})                           { |m| "spec/lib/#{m[1]}_spec.rb" }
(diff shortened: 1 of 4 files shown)


Guardfile
```diff
@@ -6,3 +6,13 @@ guard 'livereload', port: '5003', grace_period: 0.5 do
...
+
+guard :rspec, cmd: 'bundle exec rspec' do
+  watch('spec/spec_helper.rb')                        { "spec" }
+  watch('app/controllers/application_controller.rb')  { "spec/controllers" }
+  watch(%r{^spec/.+_spec\.rb$})
+  watch(%r{^app/(.+)\.rb$})                           { |m| "spec/#{m[1]}_spec.rb" }
+  watch(%r{^app/(.*)(\.erb|\.haml|\.slim)$})          { |m| "spec/#{m[1]}#{m[2]}_spec.rb" }
+  watch(%r{^lib/(.+)\.rb$})                           { |m| "spec/lib/#{m[1]}_spec.rb" }
+  watch(%r{^app/controllers/(.+)_(controller)\.rb$})  { |m| ["spec/#{m[2]}s/#{m[1]}_#{m[2]}_spec.rb", "spec/acceptance/#{m[1]}_spec.rb"] }
+end
```
damassi: Pulled this slightly-modified snippet from [the README](https://github.com/guard/guard-rspec#typical-rails-app)

spec/spec_helper.rb
```diff
@@ -3,6 +3,9 @@
 require 'yarjuf'
 
 RSpec.configure do |config|
+  config.filter_run focus: true
```
damassi: Similar to jest equivelant
```js
it.only("works", () => ... )
```

spec/spec_helper.rb
```diff
@@ -3,6 +3,9 @@
...
+  config.run_all_when_everything_filtered = true
```
damassi: Slightly strange language, but means: "when no `focus: true` filters are found, everything is filtered, which means, nothing runs, which means _run everything_"

PR #507, opened 2020-03-05
CSGN-64: Add assigned to for submissions

jonallured:
This PR adds a string column to Submission for storing which admin is working on it. I researched how to approach this work and discovered that making a connection between a Gravity Admin record and the Submission record in Convection would be a lot of work. The reason this isn't trivial is that the admin endpoint in Gravity only exists on V1 and Convection isn't currently setup to talk to that version of the API.

In order to set that up I would have had to create another Convection ClientApplication record in Gravity because those are tied to a specific version of the API. Then with that new client app, I'd also probably want to pull in Kinetic. That seemed like a lot of work so I checked with Noosh and she's totally cool with just storing this as a string for now.

It's possible that in the future we'll want something more robust, but my feeling is that we should wait until that's true and only make this connection more strict when we have to.

Screen Shot 2020-03-05 at 2 23 44 PM
Screen Shot 2020-03-05 at 2 23 59 PM
Screen Shot 2020-03-05 at 2 24 08 PM


config/initializers/_config.rb
```diff
@@ -7,6 +7,7 @@ module Convection
 Convection.config =
   OpenStruct.new(
     access_token: ENV['ACCESS_TOKEN'] || 'replace-me',
+    admin_names: (ENV['ADMIN_NAMES'] || 'Alice Betty Cindy').split,
```
damassi: I see 😄

damassi: It's worth noting that we have stitching in Gravity: https://github.com/artsy/metaphysics/tree/master/src/lib/stitching/gravity

PR #515, opened 2020-03-10
Add image urls to the asset type

jonallured:
This PR exposes the image urls on the asset type so that we can consume them over in Volt. I can't explain why these urls were not already exposed, but it's possible it was due to the lack of a client. It's also possible that we didn't want to expose them because there's an async process that produces these urls. I think that's fine for the general use case. I've demonstrated here in these specs that if the image asset has not been processed, then in the response we'll return an empty object.

That means that when a client uses this field, they are free to check for the presence of an image and if it doesn't exist do something sensible like show a placeholder. I think that's good enough and we can iterate on this if/when it makes sense to improve the flow.

Note that there's quite a bit of churn in this PR, only the final commit is all that interesting. The other two are work I took on to improve the organization of our GraphQL request specs. I've switched to organizing them by operation type: mutation and query. Further, I have a file for each of the following:

* `addAssetToConsignmentSubmission` mutation
* `createConsignmentSubmission` mutation
* `updateConsignmentSubmission` mutation
* `submissions` query

There are still missing specs here, but that goes outside the scope of this PR. I would expect for this list to grow as we evolve our GraphQL api and feel like this organization is in a better position to support that evolution.

One more thought here: I noticed a pattern where many of these request specs described how both valid and invalid setups should behave. I ended up separating them into distinct describe blocks because I found it noisy to see them together. I also worked hard to rewrite many of these specs using the features of Rspec that can help readers get the signal from the noise of a test file. I hope it helps!!!


spec/requests/api/graphql/mutations/add_asset_to_consignment_submission_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    payload = { aud: 'gravity', sub: user.gravity_user_id, roles: 'user' }
+    JWT.encode(payload, Convection.config.jwt_secret)
+  end
+
+  let(:headers) { { 'Authorization' => "Bearer #{token}" } }
+
+  let(:mutation_inputs) do
+    "{ clientMutationId: \"test\", submissionID: #{
+      submission.id
+    }, geminiToken: \"gemini-token-hash\" }"
```
jonallured: Go home prettier Ruby you're drunk. 😭

app/graph/types/asset_type.rb
```diff
@@ -9,6 +9,7 @@ module Types
       field :id, !types.ID, 'Uniq ID for this asset'
       field :asset_type, !types.String, 'type of this Asset'
       field :gemini_token, types.String, 'gemini token for asset'
+      field :image_urls, GraphQL::Types::JSON, 'known image urls'
```
damassi: One Q I have (and this came up with my `created_at` type) is the use of `GraphQL::Types` versus `types`, scoped within this file -- possible? If so, can we either update all to be `types` or all to be `GraphQL::Types`?

jonallured: Excellent question - I have this ticketed here:

https://artsyproduct.atlassian.net/browse/CSGN-83

I think we're behind other sibling services at Artsy that have already done this migration so I feel like we should be clear to work that one. I might take it one this week!

spec/requests/api/graphql/mutations/add_asset_to_consignment_submission_spec.rb
damassi: I wonder if this is due to the outer `"`, instead of a `'`? Or in ruby is the `"` required for interpolation like this?

app/graph/types/asset_type.rb
```diff
@@ -9,6 +9,7 @@ module Types
       field :id, !types.ID, 'Uniq ID for this asset'
```
damassi: One meta comment -- since we're doing some refactoring here, can we rename the folder `graph` to `graphql`? That way it matches the `graphql` folder in `specs` while also favoring the literal over the hip 😎 (No strong opinion about this, though, just something I'd rather see as a hypothetical new arrival to the codebase.)

spec/requests/api/graphql/mutations/create_consignment_submission_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+          state
+          minimum_price_dollars
+          currency
+        }
+      }
+    }
+    GRAPHQL
+  end
+
+  describe 'invalid requests' do
```
damassi: 👍 yepp

spec/requests/api/graphql/mutations/create_consignment_submission_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      body = JSON.parse(response.body)
+      create_response = body['data']['createConsignmentSubmission']
+
+      submission_response = create_response['consignmentSubmission']
+      expect(submission_response['id']).not_to be_nil
+      expect(submission_response['title']).to eq 'soup'
+      expect(submission_response['category']).to eq 'JEWELRY'
+      expect(submission_response['state']).to eq 'REJECTED'
+      expect(submission_response['minimum_price_dollars']).to eq 50_000
+      expect(submission_response['currency']).to eq 'GBP'
```
damassi: Rather than individual field testing with multiple `expect`s, is it possible to use a single expect and assert directly on the value (not instance) of a hash? Somethiing like:

```ruby
expect(response).to eq {
  title: 'soup',
  category: 'JEWELRY'
  ...
}
```

And then let the diff in the failure tell the story?

jonallured: Good call, updated! Turns out the better way to do this is:

```
expect(thingy).to include(
  key: 'value',
  another: 'value'
)
```

So that's probably better.

app/graph/types/asset_type.rb
jonallured: Sure, no strong opinion either but seems harmless!

spec/requests/api/graphql/mutations/add_asset_to_consignment_submission_spec.rb
jonallured: So this is about interp like you said. Maybe there's a better way for me to format it, but I couldn't find it. 🤷‍♂

PR #526, opened 2020-03-17
Migrate graphql code to newer DSL

jonallured:
This PR moves convection from the older style of graphql-ruby schema definition to the newer class-based style. More on that here:

https://graphql-ruby.org/schema/class_based_api.html

I took this migration in steps as seen in the commit history, so that's probably the best place to review. 

## Improving confidence

In order to be confident that I didn't break anything, I start by introducing a rake task that will detect schema drift. It dumps the schema and then compares it to the one checked into the repo. If there is a diff, then the task will fail. This has been added to the default task as well so it'll run on CI too.

Then we go and improve our request specs for the graphql endpoint. Importantly we did not have any coverage for the singular submission query so I added a file for that.

## Adding resolvers

I found it really challenging to deal with defining the schema and also resolving it in the same files so I extracted a fleet of objects called resolvers. This way the root types and mutations can define the interface with the world but then send the trio of `arguments`, `context` and `object` to a resolver class and then it's the one that knows how to validate and run the operation.

## Re-init

I found that convection was old enough to have missed some of the init goodies that a newer app would get so I ran the init task again and that dumped out a bunch of things, including a bunch of base types.

## Doing the migration

With these pieces in place I started by migrating with the included task like so:

```
$ bundle exec rake graphql:upgrade:member[app/graphql/types/asset_type.rb]
```

And then with some trial/error was able to get things to work and tests to pass. After that I did the root types and the schema itself, but that was sorta all rolled into one code fist just to get it to work. Hard to break that step down.

## Matching existing clients

At this point I posted a WIP of this work and started getting good feedback from @damassi about how this looked. He helped me see that I needed to fix a few things on the mutations because Eigen already used them. These were minor fixes, and mostly boiled down to supporting `somethingID` rather than `somethingId`. I updated my tests to match this behavior and got a final commit in for that.

Then, I pushed this new schema into my local MP and then that new schema into local Eigen and booted everything up locally. I went through the flow of creating a consignment submission and it worked!

FWIW, I plan to do this again as this stuff lands on staging.


_schema.graphql
```diff
@@ -9,7 +9,7 @@ input AddAssetToConsignmentSubmissionInput {
   """
   clientMutationId: String
   geminiToken: String!
-  submissionID: ID!
+  submissionId: ID!
```
damassi: Is this gonna blow anything up?

damassi: (`D` to `d`

jonallured: Yeah, I’m unclear about what’s safe/risky here so could use some direction. What I should probably do is "smoke test" this locally with these changes pushed into a local copy of MP. 🤔

app/graphql/mutations/create_submission_mutation.rb
```diff
@@ -1,50 +1,41 @@
...
-        input_field :locationState, types.String
-        input_field :medium, types.String
-        input_field :minimumPriceDollars, types.Int
-        input_field :provenance, types.String
-        input_field :signature, types.Boolean
-        input_field :state, Types::StateType
-        input_field :title, types.String
-        input_field :width, types.String
-        input_field :year, types.String
+    argument :additional_info, String, required: false
```
damassi: Do you know if mutations `camelCase` by default? This was an area that had to be explicitly matched between old and new schema, as this is what Eigen uses.

app/graphql/mutations/create_submission_mutation.rb
```diff
@@ -1,50 +1,41 @@
...
+    field :consignment_submission, Types::SubmissionType, null: true
 
-    def self.resolve(_obj, args, context)
-      params = args.to_h['input'].except('clientMutationId')
-      client_mutation_id = args.to_h['input']['clientMutationId']
+    def resolve(arguments)
+      resolve_options = {
+        arguments: arguments, context: context, object: object
+      }
+      resolver = CreateSubmissionResolver.new(resolve_options)
```
damassi: So this is 1:1 with `SubmissionService.create_submission`?

app/graphql/mutations/update_submission_mutation.rb
```diff
@@ -1,65 +1,42 @@
...
-
-      SubmissionService.update_submission(submission, params.except('id'))
-      OpenStruct.new(
-        consignmentSubmission: submission,
-        client_mutation_id: client_mutation_id
-      )
+  class UpdateSubmissionMutation < Mutations::BaseMutation
+    argument :id, ID, required: true
+
+    argument :additional_info, String, required: false
```
damassi: Same thing here, we'll need to be sure that mutation args are automatically camel-cased. This will likely throw an error in MP <-> Eigen

app/graphql/types/base_argument.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Types
+  class BaseArgument < GraphQL::Schema::Argument; end
+end
```
damassi: Can you explain the purpose of this and the following classes / why they need to be broken out? I would personally rather tap into `GraphQL::Schema::Enum` directly unless that type (and the following below) need to be extended significantly. Until then it seems a bit premature.

app/graphql/types/base_scalar.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Types
+  class BaseScalar < GraphQL::Schema::Scalar; end
```
damassi: Similarly, according to the docs
> The GraphQL schema language supports the scalar types of String, Int, Float, Boolean, and ID which are the most primitive kinds of values, so creating base-classes feels unnecessary.

app/graphql/types/base_union.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Types
+  class BaseUnion < GraphQL::Schema::Union; end
+end
```
damassi: A union can only ever be a union, it too is a language primitive that can't be extended in a meaningful way

jonallured:
@damassi you asked about those base types and my short response is they come from an init task and aren't really used.

My longer answer is that I think this pattern fits in with how the Rails community exposes "stub" classes to be inherited from rather than encouraging re-opening library code. A good example of this is the ApplicationRecord class or the ApplicationController. The former exists in convection but does almost nothing:

https://github.com/artsy/convection/blob/master/app/models/application_record.rb

The later exists here:

https://github.com/artsy/convection/blob/master/app/controllers/application_controller.rb

And has picked up some responsibilities that we wanted all over the routes of the application.

I would predict that these base types will be more like the ApplicationRecord class - stubs that don't really get used. They exist only because the init task generates them and then they are used to inherit up to the library code.

I looked over both exchange and vortex and those codebases also have these files and they are empty as well. This leads me to believe that there's no harm in having these files and they are just about getting things setup as the library authors intend.

app/graphql/mutations/create_submission_mutation.rb
jonallured: The responsibilities of these Resolver classes are:

* gobble up arguments, context and object and then
* throw exceptions when things are not valid
* massage inputs and then send them to services and handle shaping the response

Does that answer your question?

app/graphql/mutations/create_submission_mutation.rb
jonallured: This is a good question, I will research!

_schema.graphql
jonallured: @damassi so if I'm understanding you correctly, I think what we want is for this file to have no diff, correct? I guess I was hoping for an opportunity to switch from snake to camel case but maybe that would break the app?

damassi:
Sounds good! I don't have particularly strong opinions about it, just noticed how the empty classes somewhat clutter the `types` folder, but would rather follow idioms 👍

app/graphql/mutations/create_submission_mutation.rb
damassi: Makes sense!

_schema.graphql
damassi: Not necessarily! What we need to be concerned about is what _has been used_ in the app -- everything else can be changed. I was looking at these files in particular: 

https://github.com/artsy/eigen/tree/master/src/lib/Components/Consignments/Submission

Which contain all of the queries. 

The thinking is backward compat, but changes made to unused areas of the codebase are fair game. (We're lucky, because the surface area of this API use is small, so we can be fairly confident of what has been used.)

app/graphql/mutations/create_submission_mutation.rb
jonallured: Yes, the default behavior is to camelCase and an option exists to opt out per field.

_schema.graphql
jonallured: Ok makes total sense now, thanks!

app/graphql/mutations/create_submission_mutation.rb
```diff
@@ -2,14 +2,16 @@
...
 
     argument :additional_info, String, required: false
     argument :authenticity_certificate, Boolean, required: false
     argument :category, Types::CategoryType, required: false
     argument :currency, String, required: false
     argument :depth, String, required: false
-    argument :dimensions_metric, String, required: false
+    argument :dimensions_metric,
+             String,
+             required: false, prepare: ->(value, _context) { value.downcase }
```
jonallured: @damassi this is what is coercing to lowercase in Rails. I figured this was fine but if you have another suggestion, I'm open!

_schema.graphql
```diff
@@ -196,17 +190,17 @@ type Query {
     """
     Get all submissions with these IDs
     """
-    ids: [ID]
+    ids: [ID!]
```
damassi: We should make sure this remains the same -- no `!` (required)

damassi: But since this is only used on volt, no biggie.

jonallured:
Re: some of the final Qs, @damassi and I chatted on Slack here:

https://artsy.slack.com/archives/CS3H1K4BD/p1584567007043900

So the punchline is that we do have to add those null settings as it's a required keyword argument and the goofy [ID!] syntax doesn't actually mean that the array of ids is required, just that if the array is given, then the elements have to be IDs. Or that's how it appears to work, haha.

app/graphql/mutations/add_asset_to_consignment_submission.rb
```diff
@@ -1,43 +1,22 @@
...
+    field :asset, Types::AssetType, null: true
 
-    def self.resolve(_obj, args, context)
-      params = args.to_h['input'].except('clientMutationId')
-      client_mutation_id = args.to_h['input']['clientMutationId']
+    def resolve(arguments)
+      resolve_options = {
+        arguments: arguments, context: context, object: object
+      }
+      resolver = AddAssetToSubmissionResolver.new(resolve_options)
```
pepopowitz: 💪

app/graphql/resolvers/base_resolver.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def initialize(context:, arguments:, object:)
+    @context = context
+    @arguments = arguments
+    @object = object
+    @error = nil
+  end
+
+  def valid?
+    raise 'Implement in subclass'
```
pepopowitz: No need for action here, just a question about ruby types. In C#, we have the concept of an abstract base class, specifically to avoid things like throwing exceptions from a base class when the subclass must implement. Does such a thing exist in ruby?

jonallured: Nope what you see here is pretty common. 🤷‍♂

jonallured:
Ok looks like all feedback is addressed and this is ready to go so going to merge and then start the process of rolling this out to MP, Eigen, Volt, etc.

PR #529, opened 2020-03-19
Add support for showing only available submissions

jonallured:
This PR adds support to the submissions query for an available flag. This flag maps to the Submission scope by the same name. What it will do is filter out submissions that are not approved but also remove those that have been consigned. That way we can show a list to partners in CMS of submission that they could make offers on.

Note that this PR also drops support for the complete flag on this same query. Unless I'm missing something, this flag was not being used and also didn't make a ton of sense. It's possible this is not the last time we'll be discussing how to filter this query, but this does get us closer to releasing our work on adding submissions to CMS, so I'm good with it, but totally open to feedback!

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -173,14 +173,14 @@ type Query {
     after: String
 
     """
+    If true return only available submissions
+    """
+    available: Boolean
+
+    """
     Returns the elements in the list that come before the specified cursor.
     """
     before: String
-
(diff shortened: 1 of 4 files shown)


jonallured:
Great, thanks pals! Going to merge and kick off another round of updates. I'm starting to think about how I can script this. 😉

PR #538, opened 2020-03-23
pin artsy/pg-data-sync image to v9.6

izakp:
… to correspond with our shared Postgres deployments

Follow-up to https://github.com/artsy/docker-images/pull/67


jonallured:
LGTM 🏆

PR #541, opened 2020-03-23
CSGN-45: Add create offer mutation

jonallured:
This PR creates a new mutation that will create an offer on a consignment submission. My approach to this work was to connect my local Volt to local MP to local Convection and then iterate on getting everything working together nicely. I started with just the three required fields of submission id, partner id and commission amount. Once I was able to get that working, then I worked through the remaining optional fields.

In terms of the response, Volt can't show offers just yet, so all I'm doing is opening up an Offer type with id. That's sufficient for now, but the next step will be to make this part more realistic.


app/graphql/types/date_type.rb
```diff
@@ -2,14 +2,17 @@
 
 module Types
   class DateType < Types::BaseScalar
-    description 'Date type'
+    description 'Date in YYYY-MM-DD format'
+    DATE_FORMAT = '%Y-%m-%d'
```
pepopowitz: There are no existing mutations that use this type that would be affected by this change, right? (Or if they would be affected, they're already using this format anyway?)

spec/requests/api/graphql/mutations/create_consignment_offer_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+        expect {
+          post '/api/graphql', params: { query: mutation }, headers: headers
+        }.to change(Offer, :count).by(1)
+
+        expect(response.status).to eq 200
+        body = JSON.parse(response.body)
+        create_response = body['data']['createConsignmentOffer']
+
+        offer_response = create_response['consignmentOffer']
+        expect(offer_response).to include({ 'id' => be })
```
pepopowitz: Just a question about ruby code that I don't understand - what does this line mean? Specifically `=> be`?

spec/requests/api/graphql/mutations/create_consignment_offer_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+        }.to change(Offer, :count).by(1)
+
+        expect(response.status).to eq 200
+        body = JSON.parse(response.body)
+        create_response = body['data']['createConsignmentOffer']
+
+        offer = Offer.last
+
+        offer_response = create_response['consignmentOffer']
+        expect(offer_response).to include({ 'id' => offer.id.to_s })
```
pepopowitz: Is there a reason to check the actual ID in this test instead of doing whatever it is that we did on ln 157? No action expected, just curious if there was a reason you thought this test was worth being more specific in the assertion.

app/graphql/types/date_type.rb
jonallured: Yeah exactly, I just ripped this off of Exchange:

https://github.com/artsy/exchange/blob/master/app/graphql/types/date_type.rb

I'm not sure how this type came to be in the first place, but this way we can parse out dates in a uniform way.

spec/requests/api/graphql/mutations/create_consignment_offer_spec.rb
jonallured: Good question, I ended up replacing this with a comparison using the offer's id because that seemed more clear and would better lend itself to further expectations as we fill this out.

BUT! To your question, here's what's up - that `offer_response` is a hash. That means rspec gives us the `include` matcher. That matcher accepts an argument that will be checked only for what it includes. This means that offer response may have lots of keys, but it must at least has a key called id with a value. The last bit of magic is that `be` part - this is another magic bit from rspec and is basically saying that it's not nil. 😝

spec/requests/api/graphql/mutations/create_consignment_offer_spec.rb
jonallured: Oops, my bad! I thought I had changed both of these but did not. Will update!

PR #543, opened 2020-03-25
CSGN-104: Add offers to submission query

jonallured:
This PR adds the offers for a given partner to the submission query. I've also added a bunch of fields to the offer type so that I can properly consume it over on Volt.

This was an interesting exercise because I didn't know one could pass an argument to a field in a query. It seems obvious now in hindsight, but I assumed I'd have to pass that argument to the top-level query like I do the submission id and then use it some other way.

But nope, just have to pass it right where we're requesting the offers.

Note: I did add a spec that ensures one can not see the offers for another partner.

https://artsyproduct.atlassian.net/browse/CSGN-104

### DIFF
--- a/app/graphql/types/submission_type.rb
+++ b/app/graphql/types/submission_type.rb
@@ -31,5 +31,14 @@ module Types
     field :user_id, String, null: false
     field :width, String, null: true
     field :year, String, null: true
+
+    field :offers, [Types::OfferType], null: false do
+      argument :gravity_partner_id, ID, required: true
+    end
+
+    def offers(gravity_partner_id:)
+      partner = Partner.find_by(gravity_partner_id: gravity_partner_id)
+      object.partner_submissions.find_by(partner: partner).offers
+    end
   end
 end
--- a/spec/requests/api/graphql/queries/submission_spec.rb
+++ b/spec/requests/api/graphql/queries/submission_spec.rb
@@ -120,13 +120,10 @@ describe 'submission query' do
         <<-GRAPHQL
         query {
           submission(#{query_inputs}) {
-            id,
-            artistId,
-            title
             offers(gravityPartnerId: "#{partner.gravity_partner_id}") {
               id
               state
-              commissionPercent
+              commissionPercentWhole
             }
           }
         }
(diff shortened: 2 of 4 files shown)


spec/requests/api/graphql/queries/submission_spec.rb
```diff
@@ -102,5 +102,49 @@
...
+        <<-GRAPHQL
+        query {
+          submission(#{query_inputs}) {
+            id,
+            artistId,
+            title
+            offers(gravityPartnerId: "#{partner.gravity_partner_id}") {
+              id
+              state
+              commissionPercent
```
pepopowitz: This might be an opinion, but there are several fields in this query that don't appear in the test assertions. If they're not worth asserting in the results, IMO they shouldn't be in the query - they have the potential to throw people off from what this test is actually testing. 

(Shameless plug, I wrote a blog post about this https://www.stevenhicks.me/blog/2018/01/chekhovs-gun-and-better-unit-tests/)

app/graphql/types/submission_type.rb
```diff
@@ -31,5 +31,14 @@ class SubmissionType < Types::BaseObject
     field :user_id, String, null: false
     field :width, String, null: true
     field :year, String, null: true
+
+    field :offers, [Types::OfferType], null: false do
+      argument :gravity_partner_id, ID, required: true
```
damassi: Yeah, we often do something like this in Reaction where we request different image sizes, so something like 

```
image(size: "large")
```

app/graphql/types/submission_type.rb
```diff
@@ -31,5 +31,14 @@ class SubmissionType < Types::BaseObject
...
+    end
+
+    def offers(gravity_partner_id:)
+      partner = Partner.find_by(gravity_partner_id: gravity_partner_id)
+      object.partner_submissions.find_by(partner: partner).offers
```
damassi: What is `object`?

jonallured: In this case it's the `Submission` class. What you see here could be written as:

```ruby
Submission.partner_submissions.find_by(partner: partner).offers
```

But using `object` seems to be more in line with how this library wants one to use their DSL. Reading through these two docs helped me better understand how these field methods are supposed to work:

https://graphql-ruby.org/fields/introduction.html
https://graphql-ruby.org/fields/arguments.html

But let me know if I can help clarify anything further!!

spec/requests/api/graphql/queries/submission_spec.rb
jonallured: > If a line of setup code isn’t important to describe a specific test, then I don’t want it to distract me.

I really like that line! With this in mind going to take another pass at this file and see if I can adhere to this principle more closely! 👍

jonallured: Addressed with 3c77374 and fixed a bug along the way!

PR #547, opened 2020-03-30
Add script to for updating libraries

jonallured:
This PR adds a script to help automate the actions of creating a routine library update commit and PR. Note that I'm replacing the existing update script because I think this is better, haha.


pepopowitz:
💪

PR #548, opened 2020-03-31
CSGN-114: Switch Category to String on Queries

jonallured:
We ran into this situation where we were querying for submission information in Volt but rather than getting a simple string for category, we instead were getting an enum. My solution to this problem is to continue to accept the value as an enum in the mutation but to return it as a simple string in the submission type. This approach would require us to ensure that Eigen doesn't need the value of its mutation return, something that I would like to test locally.

https://artsyproduct.atlassian.net/browse/CSGN-114

### DIFF
--- a/app/graphql/types/submission_type.rb
+++ b/app/graphql/types/submission_type.rb
@@ -8,7 +8,7 @@ module Types
     field :artist_id, String, null: false
     field :assets, [Types::AssetType, null: true], null: true
     field :authenticity_certificate, Boolean, null: true
-    field :category, Types::CategoryType, null: true
+    field :category, String, null: true
     field :created_at, GraphQL::Types::ISO8601DateTime, null: true
     field :currency, String, null: true
     field :depth, String, null: true
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -291,7 +291,7 @@ type Submission {
   artistId: String!
   assets: [Asset]
   authenticityCertificate: Boolean
-  category: Category
+  category: String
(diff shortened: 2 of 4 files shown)


app/graphql/types/submission_type.rb
```diff
@@ -8,7 +8,7 @@ class SubmissionType < Types::BaseObject
     field :artist_id, String, null: false
     field :assets, [Types::AssetType, null: true], null: true
     field :authenticity_certificate, Boolean, null: true
-    field :category, Types::CategoryType, null: true
+    field :category, String, null: true
```
damassi: @jonallured - i believe these things are connected: 

image

jonallured: @damassi we can specify the type of the input to the mutation separately from the type of the return from the query. They are not connected in that way.

damassi: Sounds good, how all these things connect is still a bit labrythine to me so wanted to surface just in case

pepopowitz:
Makes sense - if it checks out on eigen, :shipit: !

jonallured:
Ok it's taken me a while to get back to this, but this schema change does not break anything in Eigen. To verify this I spun up my local convection and MP servers and opened the app. I connected all these things together and then I pushed this schema update all the way through the stack.

Then I made a new consignment submission and saw that it landed in the local convection database and that there were no errors in the UI of Eigen. That makes me feel confident that this change is legal and I'm going to merge and shepherd this through our projects.

PR #563, opened 2020-04-07
CSGN-126: Allow partners to see all submissions

jonallured:
This PR fixes a bug where only admin users were able to see all submissions as documented in this bug:

https://artsyproduct.atlassian.net/browse/CSGN-126

So my approach here is to take a look at the roles of the user and if they are a partner, allow them to see all submissions.

Note: there's some unrelated diff here where the RuboCop setup needed some tweaks that I hadn't noticed before in order to suppress some warnings.

### DIFF
--- a/app/graphql/resolvers/submissions_resolver.rb
+++ b/app/graphql/resolvers/submissions_resolver.rb
@@ -9,31 +9,12 @@ class SubmissionsResolver < BaseResolver
       'Only Admins can use the user_id for another user.'
     )
 
+  BadArgumentError = GraphQL::ExecutionError.new("Can't access arguments: ids")
+
   def valid?
     return true if admin?
 
-    if @arguments.key?(:ids)
-      bad_argument_error =
-        GraphQL::ExecutionError.new("Can't access arguments: ids")
-      @error = bad_argument_error
-      return
-    end
-
-    get_all_submissions = @arguments[:ids].empty? && @arguments[:user_id].empty?
-    user_mismatch =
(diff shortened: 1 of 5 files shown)


app/graphql/resolvers/submissions_resolver.rb
```diff
@@ -43,11 +26,39 @@ def run
...
     @arguments[:available] ? Submission.available : Submission.all
   end
 
   def conditions
-    { id: @arguments[:ids], user_id: @arguments[:user_id] }.compact
+    { id: submission_ids.presence, user_id: user_ids.presence }.compact
+  end
+
+  def submission_ids
+    @arguments.fetch(:ids, [])
```
damassi, line 48: What does it mean to execute `fetch` on `arguments`? Is this for providing defaults when a value is not found?

jonallured: Yep, that's exactly right! https://ruby-doc.org/core-2.5.1/Hash.html#method-i-fetch

These are (mostly) equal:

```ruby
@arguments.fetch(:ids, [])
@arguments[:ids] || []
```

Worth noting that fetch will return nils!

```ruby
irb(main):012:0> hsh = { omg: nil }
irb(main):013:0> hsh[:omg] || "lol"
=> "lol"
irb(main):014:0> hsh.fetch(:omg, "lol")
=> nil
```

So it sorta depends what you're trying to do, but I tend to reach for fetch first as I like how it reads better.

PR #568, opened 2020-04-08
[CSGN-37] Add pagination to submissions connection

pepopowitz:
Partially addresses https://artsyproduct.atlassian.net/browse/CSGN-37

This PR adds pagination to the submissions connection in the convection GraphQL schema. It is inspired by similar work [in Exchange](https://github.com/artsy/exchange/blob/master/app/graphql/types/pagination/pageable_connection.rb). These specific pagination fields are required by the [`Pagination` component in Reaction](https://github.com/artsy/reaction/blob/master/src/Components/Pagination.tsx#L53), and will make it easy for us to implement paging in Volt.

image

The implementation code is very similar to what is in Exchange, but there are some differences. @jonallured and I considered putting it in a common gem for both to consume, but "the rule of three" influenced the decision to duplicate for now. 

The biggest differences in the implementation are that the API for the graphql-ruby gem changed in the convection version (1.10). Specifically:

- `object.nodes` used to provide the _complete_ set of nodes for a pageable connection, but now it only provides the current page of nodes. There is a new property named `object.items` that I had to use instead.
- `object.encode` is no longer available. I resolved this by tracking down graphql-ruby's implementation of that function, and pulled it into our `PageableConnection`.

Other differences are in the tests. I removed some that seemed to not test anything unique, and also rewrote them a bit to jive better with our existing tests in convection. I didn't want to change them _too_ much because I think it's important if we get to a third implementation of this pagination logic to easily recognize the duplication. 

## Follow-up work coming soon

- Update the metaphysics schema to include these new fields
- Use the [`Pagination` component from Reaction](https://github.com/artsy/reaction/blob/master/src/Components/Pagination.tsx) to implement pagination of submissions in Volt.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -204,6 +204,38 @@ type Offer {
   state: String
 }
 
+type PageCursor {
+  """
+  first cursor on the page
+  """
+  cursor: String!
+
+  """
+  is this the current page?
+  """
+  isCurrent: Boolean!
+
+  """
+  page number out of totalPages
+  """
+  page: Int!
+}
+
+type PageCursors {
+  around: [PageCursor!]!
+
+  """
+  optional, may be included in field around
+  """
+  first: PageCursor
+
+  """
+  optional, may be included in field around
+  """
+  last: PageCursor
+  previous: PageCursor
+}
+
 """
 Information about pagination in a connection.
 """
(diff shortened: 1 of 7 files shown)


damassi:
Looks great @pepopowitz 👍 This pagination implementation is pretty cool and also very portable / flexible UI-wise.

PR #570, opened 2020-04-09
Fix formatting and linting errors from previous merge

pepopowitz:
#568 had some prettier errors, which showed up in a failed build. I discovered that my local version of prettier or prettier-ruby was outdated, and that's how the commit slipped through without hooks catching it. I don't know why checks in GitHub didn't fail, though - I'll dig into that. But for now, this PR should get master building again 🤞


pepopowitz:
It looks like checks didn't fail because they didn't run the tests for #568.

pepopowitz:
Oooooohhhhh because I forgot that we are PR'ing from branches in artsy/convection, not from forks. Closing this and re-opening from a branch in this repo.

PR #578, opened 2020-04-14
CSGN-141: Support publishing of submissions

jonallured:
This PR addresses a few tickets by introducing a couple new states for submission:

https://artsyproduct.atlassian.net/browse/CSGN-141
https://artsyproduct.atlassian.net/browse/CSGN-142

What was discovered through QA is that we were showing all approved submissions to partners in CMS, but we don't always want to do this. Sometimes we approve a submission and then want to share it with partners in a customized way.

To support this, we've added a state called 'published' and then updated the definition of what the available scope means to only include published submissions. This means we have complete control over which submissions show to our partners.

But then the flip side of this is that we'll want to be able to take down submissions so we've also added a 'closed' state. Moving a submission to this state will exclude it from CMS.

Note: while we were at it, we clarified that approval does NOT add the work to the daily digest either.

Ok time for some screenshots!!

<details>

Screen Shot 2020-04-14 at 3 20 32 PM
Screen Shot 2020-04-14 at 3 20 36 PM
Screen Shot 2020-04-14 at 3 20 40 PM
Screen Shot 2020-04-14 at 3 20 44 PM
Screen Shot 2020-04-14 at 3 20 58 PM
Screen Shot 2020-04-14 at 3 21 15 PM

</details>

To QA this what I did was take a submitted submission and publish/undo then close/undo to see that it went through the right state changes.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -21,11 +21,13 @@ Metrics/AbcSize:
 # ExcludedMethods: refine
 Metrics/BlockLength:
   Max: 411
+  Exclude:
+    - 'spec/**/*'
 
 # Offense count: 7
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 172
+  Max: 100000
 
 # Offense count: 1
 Metrics/CyclomaticComplexity:
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -16,6 +16,7 @@ class StateActions
     actions << approve_action if submission.submitted?
     actions << publish_action if submission.submitted? || submission.approved?
     actions << reject_action if submission.submitted?
+    actions << close_action if submission.submitted?
     actions
   end
 
@@ -54,6 +55,15 @@ class StateActions
       text: 'Reject'
     }
   end
+
+  def close_action
+    {
+      class: default_classes << 'btn-delete',
+      confirm: 'No email will be sent.',
+      state: 'closed',
+      text: 'Close'
+    }
+  end
 end
 
 module Admin
@@ -68,6 +78,7 @@ module Admin
                     undo_approval
                     undo_publish
                     undo_rejection
+                    undo_close
                   ]
 
     expose(:submissions) do
@@ -182,6 +193,11 @@ module Admin
 
     def undo_rejection
       SubmissionService.undo_rejection(@submission)
+      redirect_to admin_submission_path(@submission)
+    end
+
+    def undo_close
+      SubmissionService.undo_close(@submission)
       redirect_to admin_submission_path(@submission)
     end
 
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -1,5 +1,61 @@
 # frozen_string_literal: true
 
+class StateActions
+  def self.for(submission)
+    new(submission).run
+  end
+
+  attr_reader :submission
+
+  def initialize(submission)
+    @submission = submission
+  end
+
+  def run
+    actions = []
+    actions << approve_action if submission.submitted?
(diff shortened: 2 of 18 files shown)


.rubocop_todo.yml
```diff
@@ -21,11 +21,13 @@ Metrics/AbcSize:
...
 Metrics/BlockLength:
   Max: 411
+  Exclude:
+    - 'spec/**/*'
 
 # Offense count: 7
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 172
+  Max: 100000
```
pepopowitz, line 30: Is this big enough? 🤔

app/controllers/admin/submissions_controller.rb
```diff
@@ -1,11 +1,85 @@
...
+
+  def default_classes
+    ['btn btn-secondary btn-small btn-full-width']
+  end
+
+  def approve_action
+    {
+      class: default_classes << 'btn-approve',
+      confirm:
+        'An email will be sent to the consignor, letting them know that their submission will be sent to our partner network and this work will appear in the digests and CMS. This action cannot be undone.',
```
pepopowitz, line 33: I think this message is inaccurate now, as we no longer send emails for the `Approve` action. I think we could reuse the message from the old `Approve without digest` action: 

```suggestion
        'No email will be sent to the consignor and this submission will be excluded from the digests.',
```

app/views/admin/submissions/_state_actions.html.erb
```diff
@@ -1,39 +1,16 @@
...
-<% elsif @submission.reviewed? %>
-  <div class='overview-section'>
-    <div class='bold-label'>Internal Approval</div>
-    <div class='single-padding-top'>
-      <%= reviewer_byline(@submission) %>
-    </div>
-  </div>
-<% end %>
+    <% end %>
+    <% @actions.each do |action| %>
```
pepopowitz, line 6: 💪 This turned out really nice. ❤️

spec/services/submission_service_spec.rb
```diff
@@ -143,16 +137,31 @@
     end
 
     it 'generates partner submissions on an approval' do
```
pepopowitz, line 139:
```suggestion
    it 'does not generate partner submissions on an approval' do
```

app/controllers/admin/submissions_controller.rb
```diff
@@ -1,11 +1,85 @@
 # frozen_string_literal: true
 
+class StateActions
```
damassi, line 3: It would be nice if this was in its own file so that it's clearly discoverable via cmd+p and sidebar.

jonallured: You're totally right I was being lazy, haha. I extracted to a separate file with 3d1ceb7 and even added some specs as punishment. 😝

spec/services/submission_service_spec.rb
jonallured: Fixed with 40266b5.

app/controllers/admin/submissions_controller.rb
jonallured: Fixed with 40266b5.

app/models/submission_state_actions.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+class SubmissionStateActions
```
damassi, line 3: Not sure if this is "the ruby way", but since there are now files associated with the `submission` model, would it make sense to do 

```ruby
- models
--- submission
------ submission.rb
------ submission_stater_actions.rb
```
?

jonallured: I guess I'd probably wait on something like this but as things evolve we could certainly move to a structure like this. The thing is that with a structure like that it implies a nesting and modules that IMO are overkill. 🤷

PR #581, opened 2020-04-15
New states QA feedback

jonallured:
This PR gets a couple minor QA items from Noosh that came up today. It updates the labels of some buttons and adds the close button to the approved submission page. 

Screenshots!!! 🔫 

<details>

Screen Shot 2020-04-15 at 3 43 45 PM
Screen Shot 2020-04-15 at 3 43 55 PM

</details>

Note: how cool is it that adding the close button was as easy as updating the `SubmissionStateActions` class and then the associated spec?? 😎

PR #582, opened 2020-04-16
Add close button to published submissions page

jonallured:
Like on #581 this PR is adding a button to the submission page for a particular state. This came up on Slack while Noosh was doing some QA. Note that I switched to a case statement for this rather than conditionals because RuboCop was yelling at me but I think it's nice actually.

### DIFF
--- a/app/models/submission_state_actions.rb
+++ b/app/models/submission_state_actions.rb
@@ -12,12 +12,16 @@ class SubmissionStateActions
   end
 
   def run
-    actions = []
-    actions << approve_action if submission.submitted?
-    actions << publish_action if submission.submitted? || submission.approved?
-    actions << reject_action if submission.submitted?
-    actions << close_action if submission.submitted? || submission.approved?
-    actions
+    case submission.state
+    when Submission::SUBMITTED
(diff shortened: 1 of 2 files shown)


app/models/submission_state_actions.rb
```diff
@@ -12,12 +12,16 @@ def initialize(submission)
   end
 
   def run
-    actions = []
-    actions << approve_action if submission.submitted?
-    actions << publish_action if submission.submitted? || submission.approved?
-    actions << reject_action if submission.submitted?
-    actions << close_action if submission.submitted? || submission.approved?
-    actions
+    case submission.state
```
pepopowitz, line 15: ❤️ ❤️ ❤️

jonallured: Look at that it's like you're a Ruby programmer now!!

pepopowitz: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

PR #589, opened 2020-04-22
Don't update prior approval info

jonallured:
While working on adding publishing to convection I made a mistake and did not guard that prior approval info should be maintained. This PR fixes that mistake by adding a test that asserts that the prior info stays put and then the addition of a guard to that update.

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -100,7 +100,10 @@ class SubmissionService
     end
 
     def publish!(submission, current_user)
-      submission.update!(approved_by: current_user, approved_at: Time.now.utc)
+      unless submission.approved_at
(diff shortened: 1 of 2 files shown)


jonallured:
Fun fact! Ruby creates times with more precision than PG will store and rehydrate - this causes problems when you try to compare times like I'm doing in this spec, you can see that difference here:

https://app.circleci.com/pipelines/github/artsy/convection/907/workflows/336c8ce5-b9a7-41e2-962e-262b0a98a7e1/jobs/1799

So I updated the spec to reference the beginning of the day which means there should be no millisecond precision and the test should pass. Programming computers is so fun!!! 🎹

PR #624, opened 2020-05-05
Library Updates

jonallured:

```
$ bundle update
$ yarn upgrade --latest
```

```
$ ./bin/update
```


jonallured:
Note that I had to pin ddtrace because it's having issues:

https://github.com/DataDog/dd-trace-rb/issues/1021

PR #631, opened 2020-05-06
Remove the 2nd of 3 reminder emails when submissions lack photos.

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-152

The consignments team doesn't want us to send so many reminders for missing images, so we're removing the middle one (2 of 3). 

The cadence of reminders used to be 
- reminder 1: right away
- reminder 2: 1 day later
- reminder 3: 7 days later

The new cadence will be:
- reminder 1: right away
- reminder 2: 7 days later

There are no visual changes to this work.

---

Follow-up work (at time of deployment to production)
- Remove the `THIRD_REMINDER_DAYS_AFTER` env setting in prod
- Set the `SECOND_REMINDER_DAYS_AFTER` env setting to `7` in prod


app/mailers/user_mailer.rb
```diff
@@ -46,27 +46,12 @@ def second_upload_reminder(submission:, user:, user_detail:)
     @user_detail = user_detail
     @utm_params =
       utm_params(
-        source: 'drip-consignment-reminder-e02',
+        source: 'drip-consignment-reminder-e02-v2',
```
pepopowitz, line 49: Updating the utm-source to reflect that this second reminder is different than it used to be.

app/views/user_mailer/second_upload_reminder.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= stylesheet_link_tag 'emails', media: 'all' %>
 
-<table class='second-upload-reminder-email'>
+<table class='third-upload-reminder-email'>
```
jonallured, line 3: This surprised me - can you help me understand what's up here?

pepopowitz: I renamed `third_upload_reminder.html.erb` to `second_upload_reminder.html.erb` after deleting the original `second`, but I didn't change any of the contents. I will, though - I've changed "third" to "second" everywhere else, so this will be confusing. Good catch!

pepopowitz: Should be fixed in a059027!

jonallured:
LGTM!! 🎉

PR #637, opened 2020-05-11
Library Updates

jonallured:

```
$ bundle update
$ yarn upgrade --latest
```

```
$ ./bin/update
```


jonallured:
Note I opted to lock faraday based on some chatter here: https://github.com/codegram/hyperclient/issues/161

PR #640, opened 2020-05-11
Upgrade Ruby to latest

jonallured:
I _think_ this is all that's required to bump convection to latest Ruby but not totally sure. Let's see what CircleCI thinks of it!


jonallured:
Looks like we're good so merging!

PR #658, opened 2020-05-22
[CSGN-140] update schema to include new saleLocation argument

xtina-starr:
Follow up to https://github.com/artsy/convection/pull/656. This PR updates the schema which was needed because of the change to the [create offer mutation](https://github.com/artsy/convection/blob/master/app/graphql/mutations/create_offer_mutation.rb#L24).

Screen Shot 2020-05-22 at 2 32 24 PM

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -91,6 +91,7 @@ input CreateOfferMutationInput {
   partnerInfo: String
   photographyInfo: String
   saleDate: Date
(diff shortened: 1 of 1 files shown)


jonallured:
I'm still trying to figure out what's up with our CI setup that I can't get a broken build to show that we'll catch an omission like this in the future, but oh well let's merge this and get on with it.

PR #659, opened 2020-05-22
Run all checks using the default rake task

jonallured:
I suspect that this will fail on Circle because we actually have an issue that @xtina-starr fixes with https://github.com/artsy/convection/pull/658 so this should be red, then I can merge hers and then rebase and this should be green and prevent any further PRs from passing when the schema file is out of sync.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -26,10 +26,8 @@ if Rails.env.development? || Rails.env.test?
 
   desc 'check for schema drift'
   task schema_check: :environment do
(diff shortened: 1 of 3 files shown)


jonallured:
Ok, after much thrashing about, I think I've got this nailed down! The punchline of this PR is that CI now runs the default rake task which includes a script to check that our schema has not drifted. I'll do a follow up PR that breaks this rule to ensure that everything is setup correctly.

PR #660, opened 2020-05-22
WIP: check for schema drift

jonallured:
This is a test to ensure that #659 is properly checking for schema drift. It should fail on CI when the `bin/schema_check` script is run and diff is detected. 🤞

### DIFF
--- a/app/graphql/convection_schema.rb
+++ b/app/graphql/convection_schema.rb
@@ -5,7 +5,6 @@ class ConvectionSchema < GraphQL::Schema
   max_complexity 300
   default_max_page_size 20
 
-  mutation(Types::MutationType)
   query(Types::QueryType)
 
   use GraphQL::Pagination::Connections


jonallured:
Ok so this build correctly failed:

https://app.circleci.com/pipelines/github/artsy/convection/1138/workflows/342a50e3-e03a-4721-a831-87828f378242/jobs/1984/steps

But the output is not what I expected. What I was expecting to see is more like what I get locally when I run this:

```
jon@juggernaut:~/code/convection(dont-merge-trash-all-mutations)% bundle exec rake schema_check
Schema IDL dumped into ./_schema.graphql
Files _schema.graphql and _schema.graphql.bak differ
schema-check failed
```

But the output of the test command stops at the "Checking formatting..." line and I can't explain that. Will post this on Slack and see if I can get help from the hive mind. 🐝

jonallured:
Slack thread here: https://artsy.slack.com/archives/C013JUHKJJW/p1590181908011800

No resolution at this point, but we're thinking about tinkering with the output formatters of these tools to see if maybe they are not playing nicely. Going to close this for now. /cc @izakp

izakp:
@jonallured some links to what appear to be the offensive code in prettier:

In the `formatFiles` function that prints filenames of checked files:
```
if (isTTY()) {
  context.logger.log(filename, { newline: false });
}
```
https://github.com/prettier/prettier/blob/76b4970c2e0461d5d0fece32f72dff5916938489/src/cli/util.js#L455

And here where it prints a summary of checked files
```
// Print check summary based on expected exit code
if (context.argv.check) {
  context.logger.log(
    numberOfUnformattedFilesFound === 0
      ? "All matched files use Prettier code style!"
      : context.argv.write
      ? "Code style issues fixed in the above file(s)."
      : "Code style issues found in the above file(s). Forgot to run Prettier?"
  );
}
```

https://github.com/prettier/prettier/blob/76b4970c2e0461d5d0fece32f72dff5916938489/src/cli/util.js#L540

Finally here the function writes to the stderr stream:

```
return function (message, opts) {
  opts = { newline: true, ...opts };
  const stream = process[loggerName === "log" ? "stdout" : "stderr"];
  stream.write(message.replace(/^/gm, prefix) + (opts.newline ? "\n" : ""));
};
```

https://github.com/prettier/prettier/blob/76b4970c2e0461d5d0fece32f72dff5916938489/src/cli/util.js#L752

PR #663, opened 2020-05-27
[CSGN-140] add sale_location to offer type

xtina-starr:
Addresses: [CSGN-140](https://artsyproduct.atlassian.net/browse/CSGN-140)

This is follow up work from the [initial PR](https://github.com/artsy/convection/pull/656) to add `sale_location` to an offer. We missed adding this property to offer_type which was causing issue when we attempted to query it from volt.

### DIFF
--- a/spec/requests/api/graphql/queries/submission_spec.rb
+++ b/spec/requests/api/graphql/queries/submission_spec.rb
@@ -204,6 +204,7 @@ describe 'submission query' do
             {
               'id' => offer.id.to_s,
               'state' => offer.state,
+              'saleLocation' => offer.sale_location,
               'commissionPercentWhole' => offer.commission_percent_whole
             }
           )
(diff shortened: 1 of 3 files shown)


spec/requests/api/graphql/queries/submission_spec.rb
```diff
@@ -137,6 +137,7 @@
             offers(gravityPartnerId: "#{gravity_partner_id}") {
               id
               state
+              saleLocation
```
jonallured, line 140: I'd love to see `saleLocation` land in that last test too, something like this:

```diff
          expect(offer_response).to match(
            {
              'id' => offer.id.to_s,
              'state' => offer.state,
+             'saleLocation' => offer.sale_location,
              'commissionPercentWhole' => offer.commission_percent_whole
            }
          )
```

How does that sound?

xtina-starr:
I'm going to go ahead and merge.

PR #671, opened 2020-06-01
Support submission query sorting

jonallured:
This PR adds the ability to specify a column to sort by when quering for submissions:

```graphql
query {
  submissions(sort: "created_at") {
    edges {
      node {
        id
        created_at
      }
    }
  }
}
```

If omitted, the default sort remains DESC by `id`. Otherwise you can specify any value included in the Submission column names array:

```ruby
convection:development> Submission.column_names.sort
=> ["additional_info", "admin_receipt_sent_at", "approved_at", "approved_by", "artist_id", "artist_score", "assigned_to", "auction_score", "authenticity_certificate", "category", "consigned_partner_submission_id", "created_at", "currency", "deadline_to_sell", "deleted_at", "depth", "dimensions_metric", "edition", "edition_number", "edition_size", "ext_user_id", "height", "id", "location_city", "location_country", "location_state", "medium", "minimum_price_cents", "offers_count", "primary_image_id", "provenance", "qualified", "receipt_sent_at", "rejected_at", "rejected_by", "reminders_sent_count", "signature", "state", "title", "updated_at", "user_agent", "user_email", "user_id", "width", "year"]
```

## Should this be an ENUM?

I guess I was thinking String for this argument value because then as fields come and go, what's ultimately the source of truth about what can be sorted by is the database. I'm interested in hearing other thoughts here though! /cc @mzikherman

### DIFF
--- a/spec/requests/api/graphql/queries/submissions_spec.rb
+++ b/spec/requests/api/graphql/queries/submissions_spec.rb
@@ -355,5 +355,92 @@ describe 'submissions query' do
         expect(asset_response['imageUrls']).to eq({})
       end
     end
+
+    describe 'sorting' do
+      let!(:submission) { Fabricate :submission, id: 7, created_at: 1.day.ago }
+      let!(:middle_submission) do
+        Fabricate :submission, id: 8, created_at: 3.days.ago
+      end
+      let!(:last_submission) do
+        Fabricate :submission, id: 9, created_at: 2.days.ago
+      end
+
+      let(:query) do
+        <<-GRAPHQL
+        query {
+          submissions(#{query_inputs}) {
+            edges {
+              node {
+                id
+                createdAt
+              }
+            }
+          }
+        }
+        GRAPHQL
+      end
+
+      context 'without a sort column' do
+        let(:query_inputs) { 'sort: null' }
+
+        it 'returns the submissions sorted ascending by the id column' do
+          post '/api/graphql', params: { query: query }, headers: headers
+
+          expect(response.status).to eq 200
+          body = JSON.parse(response.body)
+
+          submissions_response = body['data']['submissions']
+          ids =
+            submissions_response['edges'].map { |edge| edge.dig('node', 'id') }
+          expect(ids).to eq %w[9 8 7]
+        end
+      end
+
+      context 'with an invalid sort column' do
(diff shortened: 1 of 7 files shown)


spec/requests/api/graphql/queries/submissions_spec.rb
```diff
@@ -355,5 +355,92 @@
...
+            }
+          }
+        }
+        GRAPHQL
+      end
+
+      context 'without a sort column' do
+        let(:query_inputs) { 'sort: null' }
+
+        it 'returns the submissions sorted ascending by the id column' do
```
pepopowitz, line 386:
```suggestion
        it 'returns the submissions sorted descending by the id column' do
```

As [indicated in the resolver](https://github.com/artsy/convection/pull/671/files#diff-b89069eeabeafab59e09aa4453c29bedR88), the default is descending by id. The results of this test also show descending by id, so I think this is just a mis-named test.

mzikherman:
> I guess I was thinking String for this argument value because then as fields come and go, what's ultimately the source of truth about what can be sorted by is the database.

My thinking was sort of different, rather, the sources of truth are the specific sorts/strings exposed to the client (like `IMPORTANCE_ASC` and `IMPORTANCE_DESC`). Those could then be implemented in an agnostic way (relative to the client use). In GraphQL typing, I would thus choose an enum over a string, since I think that's preferred for when the universe of options is small. You'd get autocomplete in the client, etc.

I also wouldn't allow sorting by ad-hoc columns in the database, but explicitly only an allow-list of columns (or enums, or whatever). To make the sort efficient you'd _probably_ want an index on those columns which are allowed to be sortable (and thus not allow them on all columns, or those without a corresponding index). This might not be a huge issue for Convection however.

jonallured:
@mzikherman thanks for chiming in, some really good points and things to think about there! 💪 

I took a shot at what it would be like to move to an enum for this with 48925b3 and I think I like it! Would be interested to hear if people think it's going too far. What I did was add an enum type called `SubmissionSortType` and then through some Ruby wizardry I have it generate a value for each column in the submissions table (both an ASC and DESC variant). Then I have a prepare hook that will parse out the column and direction and send that to the ActiveRecord order method. What do people think of this approach?

pepopowitz:
Cool! Gonna merge and we can address any other potential feedback in followups.

PR #704, opened 2020-06-26
Auct1107 submission notes

erikdstock:
This PR creates a `Note` model + table, which belongs to a user and a submission, and has a body. It also adds a section for the notes. Currently the section only appears if the submission has notes, which is never. In a followup PR we will add the ability to create a note on the page, at which point the section will probably always appear.

image

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Note < ApplicationRecord
+  validates :body, presence: true
   belongs_to :submission
   belongs_to :author, foreign_key: :created_by, class_name: "User", inverse_of: :authored_notes
 
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class Note < ApplicationRecord
+  belongs_to :submission
+  
+end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -3,11 +3,14 @@
 class Note < ApplicationRecord
   validates :body, presence: true
   belongs_to :submission
-  belongs_to :author, foreign_key: :created_by, class_name: "User", inverse_of: :authored_notes
+  belongs_to :author,
+             foreign_key: :created_by,
+             class_name: 'User',
+             inverse_of: :authored_notes
 
   def byline
-    (author ? "#{author.email} - " : '') + \
-    (created_at == updated_at ? '' : 'Updated ') + \
-    updated_at.to_formatted_s(:long)
+    (author ? "#{author.email} - " : '') +
+      (created_at == updated_at ? '' : 'Updated ') +
+      updated_at.to_formatted_s(:long)
   end
 end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -2,5 +2,5 @@
 
 class Note < ApplicationRecord
   belongs_to :submission
-  
+  belongs_to :author, foreign_key: :created_by, class_name: "User", inverse_of: :authored_notes
 end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -3,14 +3,9 @@
 class Note < ApplicationRecord
   validates :body, presence: true
   belongs_to :submission
-  belongs_to :author,
-             foreign_key: :created_by,
-             class_name: 'User',
-             inverse_of: :authored_notes
 
-  def byline
-    (author ? "#{author.email} - " : '') +
-      (created_at == updated_at ? '' : 'Updated ') +
-      updated_at.to_formatted_s(:long)
-  end
+  belongs_to :author,
+             foreign_key: :gravity_user_id,
+             class_name: 'User',
+             primary_key: :gravity_user_id
 end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -3,4 +3,10 @@
 class Note < ApplicationRecord
   belongs_to :submission
   belongs_to :author, foreign_key: :created_by, class_name: "User", inverse_of: :authored_notes
+
+  def byline
+    (author ? "#{author.email} - " : '') + \
+    (created_at == updated_at ? '' : 'Updated ') + \
+    updated_at.to_formatted_s(:long)
+  end
 end
--- a/app/views/admin/submissions/_note.html.erb
+++ b/app/views/admin/submissions/_note.html.erb
@@ -1,0 +1,6 @@
+  <div class='list-group-item list-item-note'>
+    <div class='list-group-item-info'>
+      <%= note.created_by %>
+    </div>
+    <textarea><%= note.body %></textarea>
+  </div>
--- a/app/views/admin/submissions/_note.html.erb
+++ b/app/views/admin/submissions/_note.html.erb
@@ -1,6 +1,10 @@
-  <div class='list-group-item list-item-note'>
-    <div class='list-group-item-info'>
-      <%= note.created_by %>
+  <div class='list-group-item--note'>
+    <div class='list-group-item--byline'>
+      <%= note.byline %>
     </div>
-    <textarea><%= note.body %></textarea>
+    <div class='list-group-item--body'>
+      <textarea >
+      <%= note.body %>
+    </textarea>
+    </div>
   </div>
--- a/app/views/admin/submissions/_notes_section.html.erb
+++ b/app/views/admin/submissions/_notes_section.html.erb
@@ -1,6 +1,6 @@
 <% if notes && notes.length > 0 %>
-  <div class='overview-section'>
-    <div class='overview-section-title--inline'>
+  <div class='notes-section'>
+    <div class='notes-section-title'>
       Notes
     </div>
     <%= render partial: 'admin/submissions/note', collection: @notes %>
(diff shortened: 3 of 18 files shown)


app/models/note.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Note < ApplicationRecord
+  validates :body, presence: true
+  belongs_to :submission
+  belongs_to :author, foreign_key: :created_by, class_name: "User", inverse_of: :authored_notes
```
erikdstock, line 6: rubocop made me declare `inverse_of` on only the `author` field, not the submission field, which is weird because a) we don't have a use case for a user's `authored_notes` and b) it didn't require it on the submission association above, which we _do_ need.

app/views/admin/submissions/_note.html.erb
```diff
@@ -0,0 +1,10 @@
+  <div class='list-group-item--note'>
```
erikdstock, line 1: I'm not sure i did my bem classes or used [artsy/watt](https://github.com/artsy/watt) right.

app/views/admin/submissions/_notes_section.html.erb
```diff
@@ -0,0 +1,8 @@
+<% if notes && notes.length > 0 %>
+  <div class='notes-section'>
+    <div class='notes-section-title'>
+      Notes
+    </div>
+    <%= render partial: 'admin/submissions/note', collection: @notes %>
```
erikdstock, line 6: I feel like this file could have gone inline in `show.html.erb`, but I was copying from the line above.

spec/views/admin/submissions/show.html.erb_spec.rb
```diff
@@ -177,6 +178,26 @@
...
+        page.visit "/admin/submissions/#{@submission.id}"
+      end
+
+      it 'shows a list of notes' do
+        expect(page).to have_content("Notes")
+        within(:css, '.notes-section') do
+          expect(page).to have_content('admin@art.sy', count: 3)
+          expect(page).to have_content('Note 1')
+          expect(page).to have_content('Note 2')
+          expect(page).to have_content('Note 3')
```
erikdstock, line 194: A bit rusty on my capybara, but this seemed to work.

spec/models/user_spec.rb
```diff
@@ -23,6 +23,14 @@
     end
   end
 
+  context 'authored_notes' do
+    it 'returns all authored_notes for the user' do
+      expect(user.authored_notes).to be_empty
```
erikdstock, line 28: A method I didn't need, but rubocop forced me to do it.

spec/models/note_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    note = Note.new
+    expect(note).not_to be_valid
+    expect(note.errors.messages[:body]).to eq ["can't be blank"]
+    
+    note.body = ''
+    expect(note).not_to be_valid
+    expect(note.errors.messages[:body]).to eq ["can't be blank"]
+
+    note.body = "f"
+    expect(note).to be_valid
```
erikdstock, line 28: IDK how I feel about this pile of assertions. Validating presence is pretty straightforward and so this was mostly me just confirming my understanding of it.

app/models/note.rb
yuki24: The `inverse_of` option could be quite dangerous because it's a way of making both sides of the relation circularly dependent of each other... I'd just disable the cop.

app/models/note.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def byline
+    (author ? "#{author.email} - " : '') + \
+    (created_at == updated_at ? '' : 'Updated ') + \
+    updated_at.to_formatted_s(:long)
+  end
```
yuki24, line 12: Seems like a method that should belong to a helper.

app/views/admin/submissions/show.html.erb
```diff
@@ -23,6 +23,7 @@
           </div>
           <%= render 'admin/consignments/consignment_section', consignment: @submission.consigned_partner_submission, artist: @submission.artist&.name %>
           <%= render 'admin/offers/offers_section', offers: @offers, truncated: true %>
+          <%= render 'admin/submissions/notes_section', notes: @notes, truncated: true %>
```
yuki24, line 26: This could probably be more straight forward for Rails users other than adding our own pattern here:

```erb
<% if @notes.size > 0 %>
  <div class='notes-section'>
    <div class='notes-section-title'>
      Notes
    </div>

    <%= render @notes %>
  </div>
<% end %>
```

`app/views/admin/notes/_note.html.erb`:

```erb
<div class='list-group-item--note'>
  <div class='list-group-item--byline'>
    <%= note.byline %>
  </div>
  <div class='list-group-item--body'>
    <textarea>
      <%= note.body %>
    </textarea>
  </div>
</div>
```

And remove related partials.

db/migrate/20200625153607_create_notes.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.string :created_by
```
yuki24, line 6: I'd use the actual primary key of the `users` table to represent relations or drop the primary key column entirely in favor of the `gravity_user_id`.

spec/models/note_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+RSpec.describe Note, type: :model do
+  include ActiveSupport::Testing::TimeHelpers
+  let(:author){Fabricate(:user, email: "admin@art.sy")}
+  let(:submission){Fabricate(:submission)}
```
yuki24, line 7: It seems like this file should be [prettified](https://github.com/prettier/plugin-ruby).

spec/models/note_spec.rb
yuki24: I wouldn't test Rails validations that do not involve any conditionals.

spec/models/user_spec.rb
yuki24: Every time Rubocop suggests something you know you don't need, that's an opportunity to disable it. Although learning about the cop itself about why it exists and how it solves a particular problem is also worth taking the time.

spec/views/admin/submissions/show.html.erb_spec.rb
yuki24: Have you tried something like:

```rb
expect(page).to have_content('Note 1\nNote 2\nNote 3')
```

spec/models/note_spec.rb
erikdstock: yeah, i just realized this was available. for some reason the format option in vscode (which points to rubocop) did not seem to be working.

db/migrate/20200625153607_create_notes.rb
erikdstock: Good points. I will reference the `gravity_user_id` field.

spec/models/user_spec.rb
erikdstock: Alright, I am removing the user.authored_notes and disabling the cop for Note#author's `belongs_to` method.

spec/views/admin/submissions/show.html.erb_spec.rb
erikdstock: This actually doesn't work because there is an author section on each note.

app/views/admin/submissions/show.html.erb
erikdstock: Awesome, did not know this was possible

spec/views/admin/submissions/show.html.erb_spec.rb
```diff
@@ -182,11 +182,9 @@
       before do
         @admin = Fabricate(:user, email: 'admin@art.sy')
         3.times do |i|
-          Fabricate(
-            :note,
-            submission: @submission,
-            created_by: @admin.id,
-            body: "Note #{i + 1}"
+          # @submission.notes.create(author: @admin, body: "Note #{i + 1}")
```
erikdstock, line 185:
```suggestion

```

app/models/note.rb
dleve123: FWIW, I'm not sure if `dependent: :nullify` is what we want here. I recall discussing (but can't find where we wrote down), that if a User A creates Notes B and C, and then User A gets deleted, we'll likely still want the orphaned foreign key on Notes B and C present, so that it's possible to at least know the shadow of a User (instead of loosing all attribution of a Note to a User).

db/schema.rb
```diff
@@ -32,6 +32,15 @@
     t.index %w[submission_id], name: 'index_assets_on_submission_id'
   end
 
+  create_table 'notes', force: :cascade do |t|
+    t.string 'gravity_user_id'
```
dleve123, line 36: Looks like it's possible that you didn't re-dump the schema after editing the migration above potentially? Considering https://github.com/artsy/convection/pull/704/files#diff-f9057bc00c8c8baad7414f865649cac2R6

I would expect this line to read:

```ruby
t.string 'gravity_user_id, null: false
```

erikdstock: Hmm, not sure how this happened but i did a `rake db:migrate:redo` and that fixed it. it is now in #707 and --i assume-- will fix itself there without the need for any other rollbacks or migrations.

PR #707, opened 2020-06-29
Create notes on a submission

erikdstock:
This PR adds the ability to create notes on a submission. I think it is pretty straightforward rails stuff but am a little rusty so definitely open to feedback on how I can do things better. For that reason I requested 3 reviewers but I'll be happy to merge with less.
Jira ticket [AUCT-1106](https://artsyproduct.atlassian.net/browse/AUCT-1106)

### DIFF
--- a/app/controllers/admin/notes_controller.rb
+++ b/app/controllers/admin/notes_controller.rb
@@ -3,8 +3,8 @@
 module Admin
   class NotesController < ApplicationController
     def create
-      @submission = Submission.find(note_params[:submission_id]) # require 'pry'; binding.pry
-      note = Note.create(gravity_user_id: @current_user, **note_params) # This doesn't actually seem to render in our layout.
+      @submission = Submission.find(note_params[:submission_id])
+      note = Note.create(gravity_user_id: @current_user, **note_params)
       unless note.persisted?
         flash[:notice] =
           "Could not create note: #{note.errors.full_messages.join(', ')}"
--- a/app/controllers/admin/notes_controller.rb
+++ b/app/controllers/admin/notes_controller.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Admin
+  class NotesController < ApplicationController
+    def create
+      @submission = Submission.find(note_params[:submission_id]) # require 'pry'; binding.pry
+      note = Note.create(gravity_user_id: @current_user, **note_params) # This doesn't actually seem to render in our layout.
+      unless note.persisted?
+        flash[:notice] =
+          "Could not create note: #{note.errors.full_messages.join(', ')}"
+      end
+
+      redirect_to admin_submission_path(@submission)
+    end
+
+    private
+
+    def note_params
+      params.require(:note).permit(:submission_id, :body)
+    end
+  end
+end
--- a/app/controllers/admin/notes_controller.rb
+++ b/app/controllers/admin/notes_controller.rb
@@ -4,13 +4,18 @@ module Admin
   class NotesController < ApplicationController
     def create
       @submission = Submission.find(note_params[:submission_id])
-      note = Note.create(gravity_user_id: @current_user, **note_params)
-      unless note.persisted?
-        flash[:notice] =
-          "Could not create note: #{note.errors.full_messages.join(', ')}"
-      end
+      note = Note.new(gravity_user_id: @current_user, **note_params)
 
-      redirect_to admin_submission_path(@submission)
+      if note.save
+        redirect_to admin_submission_path(@submission),
+                    notice: 'Note has successfully been created.'
+      else
+        redirect_to admin_submission_path(@submission),
+                    alert:
+                      "Could not create note: #{
+                        note.errors.full_messages.join(', ')
+                      }"
+      end
     end
 
     private
--- a/app/helpers/submissions_helper.rb
+++ b/app/helpers/submissions_helper.rb
@@ -87,10 +87,14 @@ module SubmissionsHelper
   end
 
   def note_byline(note)
-    (note.author ? "#{note.author.email} - " : '') +
-      (note.created_at == note.updated_at ? '' : 'Updated ') +
+    return '' if note.created_at.blank?
+
+    byline = (note.author ? "#{note.author.email} - " : '')
+    byline += 'Updated ' if note.created_at == note.updated_at
+    byline +=
       note.updated_at.in_time_zone('Eastern Time (US & Canada)').to_formatted_s(
         :long
       )
+    byline
   end
 end
--- a/app/helpers/submissions_helper.rb
+++ b/app/helpers/submissions_helper.rb
@@ -89,8 +89,8 @@ module SubmissionsHelper
   def note_byline(note)
     return '' if note.created_at.blank?
 
-    byline = (note.author ? "#{note.author.email} - " : '')
-    byline += 'Updated ' if note.created_at == note.updated_at
+    byline = (note.author ? "#{note.author.email} - " : 'User deleted - ')
+    byline += 'Updated ' if note.created_at < note.updated_at
     byline +=
       note.updated_at.in_time_zone('Eastern Time (US & Canada)').to_formatted_s(
         :long
--- a/app/views/admin/notes/_form.html.erb
+++ b/app/views/admin/notes/_form.html.erb
@@ -1,0 +1,5 @@
+<%= form_with model: [:admin, note], local: true do |f| %>
+  <%= f.hidden_field :submission_id %>
+  <%= f.text_area :body %>
+  <%= f.submit 'Create' %>
+<% end %>
--- a/app/views/admin/notes/_form.html.erb
+++ b/app/views/admin/notes/_form.html.erb
@@ -1,5 +1,5 @@
 <%= form_with model: [:admin, note], local: true do |f| %>
   <%= f.hidden_field :submission_id %>
-  <%= f.text_area :body %>
+  <%= f.text_area :body, required: "required" %>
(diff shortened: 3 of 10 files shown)


app/views/admin/submissions/show.html.erb
```diff
@@ -23,14 +23,16 @@
...
           <%= render 'admin/consignments/consignment_section', consignment: @submission.consigned_partner_submission, artist: @submission.artist&.name %>
           <%= render 'admin/offers/offers_section', offers: @offers, truncated: true %>
-          <% if @notes.size > 0 %>
             <div class='notes-section'>
               <div class='notes-section-title'>
                 Notes
               </div>
               <%= render @notes %>
+              <fieldset class='new-note'>
+                <legend>New note</legend>
```
erikdstock, line 32: not sure if it is semantically correct to wrap a form in a fieldset.

spec/views/admin/submissions/show.html.erb_spec.rb
```diff
@@ -198,6 +196,37 @@
           expect(page).to have_content('Note 3')
         end
       end
+
+      it 'shows a notes section even if there are no notes' do
+        page.visit "/admin/submissions/#{@submission.id}"
+
```
erikdstock, line 202: probably an unnecessary test - i wrote this initially because i was undoing a previous assertion above where i removed L50.

app/controllers/admin/notes_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Admin
+  class NotesController < ApplicationController
+    def create
+      @submission = Submission.find(note_params[:submission_id]) # require 'pry'; binding.pry
```
bhoggard, line 6: I would remove the comment.

app/controllers/admin/notes_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      note = Note.create(gravity_user_id: @current_user, **note_params) # This doesn't actually seem to render in our layout.
```
bhoggard, line 7: What do you mean it doesn't render?

erikdstock: ugh, these comments seem to have been formatted like so for prettier and i missed them. the 'doesnt seem to render' one is wrong, i needed to use `flash[:notice]` - `[:error]` was not rendering.

app/views/admin/submissions/show.html.erb
yuki24: What's the reason why you wanted to wrap the entire form with a `<fieldset>`?

app/views/admin/notes/_form.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= form_with model: [:admin, note], local: true do |f| %>
+  <%= f.hidden_field :submission_id %>
+  <%= f.text_area :body %>
```
yuki24, line 3: Let's add a `required` attribute so the browser will prevent the user from submitting with empty data:

```suggestion
  <%= f.text_area :body, required: "required" %>
```

app/views/admin/notes/_form.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= form_with model: [:admin, note], local: true do |f| %>
+  <%= f.hidden_field :submission_id %>
```
yuki24, line 2: I think this should be part of the path. The [shallow nesting](https://guides.rubyonrails.org/routing.html#shallow-nesting) is the Rails term for this kind of stuff.

app/controllers/admin/notes_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  class NotesController < ApplicationController
+    def create
+      @submission = Submission.find(note_params[:submission_id])
+      note = Note.create(gravity_user_id: @current_user, **note_params)
+      unless note.persisted?
+        flash[:notice] =
+          "Could not create note: #{note.errors.full_messages.join(', ')}"
+      end
+
+      redirect_to admin_submission_path(@submission)
```
yuki24, line 13: I think it is more informative to use the `notice` flash when the request is successful and use the `error` flash when it fails to create one.

```ruby
if note.save
  redirect_to admin_submission_path(@submission), notice: "Note has successfully been created."
else
  redirect_to admin_submission_path(@submission), error: "Could not create note: #{note.errors.full_messages.join(', ')}"
end
```

app/controllers/admin/notes_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Admin
+  class NotesController < ApplicationController
+    def create
+      @submission = Submission.find(note_params[:submission_id])
+      note = Note.create(gravity_user_id: @current_user, **note_params)
```
yuki24, line 7: I think it should be `author`, not `gravity_user_id`?

```suggestion
      note = Note.create(author: @current_user, **note_params)
```

app/controllers/admin/notes_controller.rb
erikdstock: there doesn't seem to be an `error` flash on the layout though.

app/controllers/admin/notes_controller.rb
erikdstock: `@current_user` is actually the gravity id though, not the instantiated object.

app/controllers/admin/notes_controller.rb
yuki24: seems like `alert` was the right one:

```ruby
ApplicationController._flash_types
# => [:alert, :notice]
```

app/controllers/admin/notes_controller.rb
yuki24: That's odd... but let's keep the `gravity_user_id` then.

spec/views/admin/submissions/show.html.erb_spec.rb
yuki24: Agreed that the test below would cover the case we want to test here.

app/views/admin/notes/_form.html.erb
erikdstock: added to PR

yuki24:
LGMT, feel free to merge once the build is ✅

spec/views/admin/submissions/show.html.erb_spec.rb
jonallured: I agree - let's remove this test for now.

spec/views/admin/submissions/show.html.erb_spec.rb
```diff
@@ -198,6 +196,37 @@
...
+      context 'creating a new note' do
+        it 'user can create a new note' do
+          page.visit "/admin/submissions/#{@submission.id}"
+
+          within(:css, '.notes-section') do
+            fill_in('note[body]', with: 'This is a really cool artwork. Wow!')
+            click_button 'Create'
+          end
+
+          within(:css, '.notes-section .list-group-item--body p') do
```
jonallured, line 215: Nice scoping!! 😍

erikdstock:
To clarify i discussed this with yuki and we decided to merge the code as it exists and follow up later to address the question of nested routes since the tests are in place now.

app/helpers/submissions_helper.rb
```diff
@@ -87,10 +87,14 @@ def formatted_minimum_price_for_email(submission)
   end
 
   def note_byline(note)
```
jonallured, line 89: I'm late with this feedback, but I would not mind this method moving to the Note class.

app/views/admin/submissions/show.html.erb
```diff
@@ -23,14 +23,16 @@
           </div>
           <%= render 'admin/consignments/consignment_section', consignment: @submission.consigned_partner_submission, artist: @submission.artist&.name %>
           <%= render 'admin/offers/offers_section', offers: @offers, truncated: true %>
-          <% if @notes.size > 0 %>
```
jonallured, line 26: This indent looks wrong to me. I think the conditional was removed, so I'd expect for the following block to be outdented, no? #minor

db/schema.rb
```diff
@@ -33,7 +33,7 @@
   end
 
   create_table 'notes', force: :cascade do |t|
-    t.string 'gravity_user_id'
+    t.string 'gravity_user_id', null: false
```
jonallured, line 36: @erikdstock do you know what's up here? I don't see a place where we added a migration to mark this is null false so maybe it was a mistake?

app/controllers/admin/notes_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      if note.save
+        redirect_to admin_submission_path(@submission),
+                    notice: 'Note has successfully been created.'
+      else
+        redirect_to admin_submission_path(@submission),
+                    alert:
+                      "Could not create note: #{
+                        note.errors.full_messages.join(', ')
+                      }"
+      end
```
jonallured, line 18: Some refactoring ideas for you:

```suggestion
      note_attrs = note_params.merge(gravity_user_id: current_user)
      note = Note.new(note_attrs)

       path = admin_submission_path(note.submission)

       if note.save
         redirect_to path, notice: 'Note has successfully been created.'
      else
        error_messages = "Could not create note: #{note.errors.full_messages.join(', ')}"
        redirect_to path, alert: error_messages
      end
```

Take or leave any/all of these ideas! 🤗

db/schema.rb
erikdstock: It was a mistake in #704 - I made it null false there but forgot to re-run the migration to regenerate the schema.

PR #709, opened 2020-06-30
respond to feedback from #707

erikdstock:
A followup to @jonallured 's [review](https://github.com/artsy/convection/pull/707#pullrequestreview-440309440).

This includes a significant change to the way we load a note's author, no longer [incorrectly] assuming admin users would be available in the local users table. As a result we call gravity for author information, currently only needed from the `SubmissionsHelper#note_byline`.


app/controllers/admin/notes_controller.rb
```diff
@@ -3,14 +3,14 @@
 module Admin
   class NotesController < ApplicationController
     def create
-      @submission = Submission.find(note_params[:submission_id])
-      note = Note.new(gravity_user_id: @current_user, **note_params)
+      note_attrs = note_params.merge(gravity_user_id: @current_user)
+      note = Note.new(note_attrs)
+      path = admin_submission_path(note.submission)
```
erikdstock, line 8: is there any case where `note.submission` could fail in a way that `Submission.find(id)` would not have above?

jonallured: If there is, let's get a test in for it!! 😍 But I don't know of one.

app/models/note.rb
```diff
@@ -4,8 +4,21 @@ class Note < ApplicationRecord
   validates :body, presence: true
   belongs_to :submission
 
-  belongs_to :author,
-             foreign_key: :gravity_user_id,
-             class_name: 'User',
-             primary_key: :gravity_user_id
+  attr_reader :author
+
+  after_initialize :set_author, if: :persisted?
```
erikdstock, line 9: Not sure if this is the best way to go about things. basically every time we fetch a user (but not after create since that is going to trigger a redirect) we want to fetch their gravity user so we can print a byline,

app/controllers/admin/notes_controller.rb
yuki24: Are we not loading the parent `submission`? It's important to respect the hierarchy otherwise it could create edge cases we may not have to cover otherwise:

```ruby
before_action :load_submission, only: :create

def create
  note = @submission.notes.new(note_params)
  path = admin_submission_path(@submission)

  if note.save
    ...
  else
    ...
  end
end

def load_submission
  # This could be inlined in #create
  @submission = Submission.find(params.dig(:note, :submission_id))
end

def note_params
  params.require(:note).permit(:submission_id, :body).merge(gravity_user_id: @current_user)
end
```

app/models/note.rb
yuki24: This logic should live in the `user` class and there is already a method that does that: https://github.com/artsy/convection/blob/fd8592f9ba4d2c4a323bb0d7b1489a70691f51ca/app/models/user.rb#L12-L16

So I think it should be a delegation so the amount of extra code would be very minimal and we can consistently use the same logic that does the same thing:

```ruby
delegate :gravity_user, to: :author
```

Also off-topic and not a blocker (but feel free to do so if you can do this quickly), but the result of `User#gravity_user` should probably be memoized to avoid making extra calls to Gravity.

app/models/note.rb
```diff
@@ -4,8 +4,21 @@ class Note < ApplicationRecord
   validates :body, presence: true
   belongs_to :submission
 
-  belongs_to :author,
-             foreign_key: :gravity_user_id,
-             class_name: 'User',
-             primary_key: :gravity_user_id
+  attr_reader :author
+
+  after_initialize :set_author
```
joeyAghion, line 9: It's pretty unconventional to request external data just to initialize a model object. Initializing model objects happens throughout tests and production operation where a complete Gravity response might not be necessary. Instead, I suggest adding a helper method (like `get_author`) to the model that lazily loads and memoizes the external data.

app/models/note.rb
erikdstock: The reason for this change is that I learned that admin users do not have a local user record, so we are going straight to gravity for their information.

app/controllers/admin/notes_controller.rb
jonallured: Yeah, that's even better, good call @yuki24! 👍

app/models/note.rb
jonallured: If I had it to do over again, I'd rename the User in this project to Consignor because that's really what it is.

app/models/note.rb
```diff
@@ -4,8 +4,17 @@ class Note < ApplicationRecord
   validates :body, presence: true
   belongs_to :submission
 
-  belongs_to :author,
-             foreign_key: :gravity_user_id,
-             class_name: 'User',
-             primary_key: :gravity_user_id
+  def author
+    @author ||= load_author
```
jonallured, line 8: I like this! So we have a lazy-loaded, memo-ized author, right? 😎

app/models/user.rb
```diff
@@ -10,9 +10,7 @@ class User < ApplicationRecord
   pg_search_scope :search, against: :email, using: { tsearch: { prefix: true } }
 
   def gravity_user
-    Gravity.client.user(id: gravity_user_id)._get
-  rescue Faraday::ResourceNotFound
-    nil
+    @gravity_user ||= load_gravity_user
```
jonallured, line 13: 💪

app/controllers/admin/notes_controller.rb
```diff
@@ -21,7 +21,7 @@ def create
     private
 
     def note_params
-      params.require(:note).permit(:submission_id, :body)
+      params.require(:note).permit(:body).merge(gravity_user_id: @current_user)
```
jonallured, line 24: Nice one!! 💪

app/models/note.rb
erikdstock: yeah, except it doesn't work because a missing `@author` will always be `nil`, so we have to check with `defined?(@author)` which is much more verbose - cc @yuki24

app/models/user.rb
```diff
@@ -10,7 +10,19 @@ class User < ApplicationRecord
   pg_search_scope :search, against: :email, using: { tsearch: { prefix: true } }
 
   def gravity_user
-    @gravity_user ||= load_gravity_user
```
jonallured: I'm not understanding why we can't use this line - can we clarify why we need to use `defined?`? /cc @yuki24

erikdstock: The reason is that the user may be `nil` which would pass through the `||=` test every time. It's easiest to see in the console: 
```
[1] pry(main)> @foo
=> nil
[2] pry(main)> defined?(@foo)
=> nil
[3] pry(main)> @foo = nil
=> nil
[4] pry(main)> defined?(@foo)
=> "instance-variable"
```

erikdstock: Yuki did point out a couple other options for this, including the [memoist](https://github.com/matthewrudy/memoist) gem.

jonallured: Ohhhh, you're saying that the return from `load_gravity_user` could be nil and if so, then we wouldn't want to run it again, we'd want to use that nil. Hrm, that stinks, haha.

jonallured: Would be cool to use the null object pattern here but probably not worth further complicating this code. 🤷

PR #721, opened 2020-07-17
CSGN-278: Allow submission edits to not affect assigned user

jonallured:
This PR filters out the `assigned_to` param when it is an empty string. This allows one to edit a submission without affecting which user the submission is assigned to. Shout out to @ashleyjelks for hanging out while I worked on this the other day!! 🙌 

https://artsyproduct.atlassian.net/browse/CSGN-278

### DIFF
--- a/spec/views/admin/submissions/edit.html.erb_spec.rb
+++ b/spec/views/admin/submissions/edit.html.erb_spec.rb
@@ -44,5 +44,12 @@ describe 'admin/submissions/edit.html.erb', type: :feature do
       expect(page).to have_content('Gob Bluth'.upcase)
       expect(page).to have_content('Add New')
     end
+
+    it 'lets you update the submission and not affect the assigned user' do
+      expect(@submission.assigned_to).to eq nil
+      find_button('Save').click
+      @submission.reload
+      expect(@submission.assigned_to).to eq nil
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/views/admin/submissions/edit.html.erb_spec.rb
```diff
@@ -44,5 +44,12 @@
       expect(page).to have_content('Gob Bluth'.upcase)
       expect(page).to have_content('Add New')
     end
+
+    it 'lets you update the submission and not affect the assigned user' do
+      expect(@submission.assigned_to).to eq nil
+      find_button('Save').click
```
pepopowitz, line 50: **thought:** This is a hard thing for me to describe, so if I don't succeed, maybe we can talk over a slack call. My understanding of this test is that it's attempting to prove that when I save an unassigned submission, it saves but stays unassigned. I don't think the steps here are testing all of that, though. In particular, I think it's missing an assertion that the save was successful.

Without that assertion that the save was successful, this test wouldn't guarantee me that the `assigned_to` hadn't changed because we allow updates without choosing an `assigned_to` instead of the save flat out failing (and _nothing_ getting updated).

pepopowitz: To be more explicit about why I care about this: without that assertion that the save was successful, it's possible we'd have false green.

jonallured: We ended up chatting a bit about this and I addressed it with 2bd715d - does that look sufficient @pepopowitz?

jonallured:
Going to merge this so I can get it out in a deploy a bit later today - lmk if you have any follow up requests @pepopowitz!

spec/views/admin/submissions/edit.html.erb_spec.rb
pepopowitz: ❤️ ❤️ ❤️

PR #732, opened 2020-07-29
Sort SubmissionSort in GraphiQL

damassi:
Adds a simple `sort` so that things are more readable in GraphiQL:

Screen Shot 2020-07-29 at 2 48 23 PM


damassi:
(Going to merge as this seems pretty uncontroversial)

PR #735, opened 2020-07-30
Remove unused nginx container and load-balancer service

joeyAghion:
This follows up https://github.com/artsy/convection/pull/734 by removing the now-unused load-balancer service and nginx container. DNS is updated for both staging and production, so by the time this is released (~tomorrow), no traffic should be flowing through the old ELB.

Detailed docs: https://www.notion.so/artsy/Nginx-Ingress-Controllers-4b5d937f0acd480f938c34ef1509d3c4

https://artsyproduct.atlassian.net/browse/PLATFORM-2590

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -61,37 +61,6 @@ spec:
             preStop:
               exec:
                 command: ["sh", "-c", "sleep 10"]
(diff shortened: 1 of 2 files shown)


damassi:
👍

PR #737, opened 2020-07-30
Add #filterByCategory arg to submissions field

damassi:
Related: https://github.com/artsy/volt/pull/4488/files 

Updates the submission field to accept a new argument `filterByCategory` --which, filters by category. 

I opted for this more specific name over, say, `filter`, because we might want to filter in a more sophisticated way in the future and appending + deprecating seems better than potentially breaking consumers. 

```graphql
{
  submissions(sort:CATEGORY_ASC, filterByCategory: PAINTING) {
    edges {
      node {
        category
        id 
        title
        createdAt
      }
    }
  }
}
```

```json
{
  "data": {
    "submissions": {
      "edges": [
        {
          "node": {
            "category": "Painting",
            "id": "1",
            "title": "sarah's test artwork!",
            "createdAt": "2017-06-06T21:58:00Z"
          }
        },
        {
          "node": {
            "category": "Painting",
            "id": "2",
            "title": "sarah's test 2",
            "createdAt": "2017-06-06T22:00:57Z"
          }
        },
        ...
     ]
   }
 }
```

#### Todo: 
- [x] Add tests

### DIFF
--- a/app/graphql/resolvers/submissions_resolver.rb
+++ b/app/graphql/resolvers/submissions_resolver.rb
@@ -43,7 +43,11 @@ class SubmissionsResolver < BaseResolver
   end
 
   def conditions
-    { id: submission_ids.presence, user_id: user_ids.presence }.compact
+    {
+      id: submission_ids.presence,
+      user_id: user_ids.presence,
+      category: filter_by_category.presence
+    }.compact
   end
 
   def submission_ids
@@ -52,6 +56,10 @@ class SubmissionsResolver < BaseResolver
 
   def user_ids
     @arguments.fetch(:user_id, [])
+  end
+
+  def filter_by_category
+    @arguments.fetch(:filter_by_category, [])
   end
 
   def not_allowed_ids?
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -753,6 +753,11 @@ type Query {
     before: String
 
     """
+    Get submissions filtered by category
+    """
+    filterByCategory: Category
+
+    """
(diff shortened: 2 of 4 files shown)


app/graphql/resolvers/submissions_resolver.rb
```diff
@@ -54,6 +58,10 @@ def user_ids
     @arguments.fetch(:user_id, [])
   end
 
+  def filter_by_category
+    @arguments.fetch(:filter_by_category, [])
```
jonallured, line 62: I'm wondering if there's a bug here when one does not supply the `filter_by_category` argument. The purpose of the `compact` above is to remove conditions that don't apply, but an empty array is not affected:

```rb
irb(main):001:0> { category: [] }.compact
=> {:category=>[]}
```

So, I'm concerned that if the `fetch` here returned an empty array that we'd inadvertently filter to only submissions with NO CATEGORY! Does that make sense? Happy to hop on a call to kick the tires and maybe you've already thought of this case, but reading through this my spidey sense was triggered, haha! 🕷️

damassi: I tried the query in all kinds of ways and didn't see a problem, though I could also be misundestanding; here's a couple screenshots: 

Screen Shot 2020-07-31 at 10 18 55 AM

Screen Shot 2020-07-31 at 10 20 22 AM

jonallured: Thanks for humoring me!! 🤗

PR #748, opened 2020-08-07
[Graphql] Update connection totalCount to return only published

damassi:
Right now we return everything, for every state. This narrows the scope a bit and will further narrow depending on which arguments are passed to the connection. 

Screen Shot 2020-08-07 at 4 54 29 PM

Screen Shot 2020-08-07 at 4 54 38 PM


damassi:
Actually, I think what we need here is the `available: true` property, and either way this approach feels a bit meh...

```graphql
{
  submissions(available:true) {
    totalCount
  }
}
```

PR #760, opened 2020-08-18
CSGN-321: Filter submissions by assigned to

jonallured:
This PR adds support for a second sort on the submissions list page:

Screen Shot 2020-08-18 at 11 54 28 AM

The values for that dropdown match up with our list of possible assignees and then includes two more:

* all => include all submissions, assigned and unassigned
* none => include only submissions without an assignee
* [name] => include only submissions for that assignee

## Extract class

My first move was to realize that the code to return the list of submissions had grown too big for the controller. I extracted a class called `SubmissionMatch` to encapsulate this logic. It takes the params as an input and returns an ActiveRecord relation that can be chained with pagination calls. I first did the extraction as a straight copy/paste but then I added unit tests and refactored it so that it would be easier to add the support for filtering by assignee.

Note: I'm def interested in naming input! Do you have a better name for the `SubmissionMatch` class? I'm all ears!! 👂 👂 👂 

## Add behavior, update UI

So with my refactored class, I then was able to add the support for the second filtering and then update the UI. I asked for some design direction and got this as a comp:

https://www.figma.com/file/hk8hC98BhKuWTdxQpBiDoY/Convection?node-id=0%3A1

I had to do a little faffing to get the watt-based layout to behave, but it wasn't too bad.

/cc @artsy/csgn-devs 

https://artsyproduct.atlassian.net/browse/CSGN-321

### DIFF
--- a/app/models/submission_match.rb
+++ b/app/models/submission_match.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class SubmissionMatch
+  def self.find_all(params)
+    matching_submissions = Submission.not_deleted
+
+    if params[:term].present?
+      matching_submissions = matching_submissions.search(params[:term])
+    end
+
+    if params[:state].present?
+      matching_submissions = matching_submissions.where(state: params[:state])
+    end
+
+    if params[:user].present?
+      matching_submissions = matching_submissions.where(user_id: params[:user])
+    end
+
+    sort = params[:sort].presence || 'id'
+    direction = params[:direction].presence || 'desc'
+
+    if sort.include?('users')
+      matching_submissions.includes(:user).reorder(
+        "#{sort} #{direction}, submissions.id desc"
+      )
+    else
+      matching_submissions.reorder("#{sort} #{direction}")
+    end
+  end
+end
--- a/app/models/submission_match.rb
+++ b/app/models/submission_match.rb
@@ -2,29 +2,49 @@
 
 class SubmissionMatch
   def self.find_all(params)
-    matching_submissions = Submission.not_deleted
+    new(params).find_all
+  end
 
-    if params[:term].present?
-      matching_submissions = matching_submissions.search(params[:term])
-    end
+  attr_reader :params
 
-    if params[:state].present?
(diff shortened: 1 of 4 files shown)


spec/models/submission_match_spec.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe SubmissionMatch do
```
pepopowitz, line 5: **praise:**  Strong test coverage. 💪 

_Comment format based on [conventionalcomments.org](https://conventionalcomments.org/)_

app/models/submission_match.rb
```diff
@@ -25,7 +25,16 @@ def term
   end
 
   def query
-    { state: params[:state].presence, user_id: params[:user].presence }.compact
+    attributes = {
+      state: params[:state].presence,
+      user_id: params[:user].presence
+    }.compact
+    attributes.merge!(assigned_to: assigned_to) if filtering_by_assigned_to?
```
pepopowitz, line 32: **question [I'm learning]:** 

Is there a construct in ruby that would allow us to inline the toggling of the `assigned_to` attribute in the `attributes` assignment? I know this doesn't work, but something like....

```
    attributes = {
      state: params[:state].presence,
      user_id: params[:user].presence,
      assigned_to: assigned_to if filtering_by_assigned_to?
    }.compact
```

_Comment format based on [conventionalcomments.org](https://conventionalcomments.org/)_

pepopowitz:
Reassigning back to you @jonallured to merge after figuring out the broken test situation.

app/models/submission_match.rb
jonallured: Here's the rub: we want to conditionally suppress a key from this hash. I'm not aware of another way other than what you see here. Let's revisit our possible choices in the UI:

* all => suppress the assigned_to key from the hash altogether
* none => add the assigned_to key with a value of nil so you can find the ones that are unassigned
* alice (or whatever) => return only the submissions with a matching assignee

So it's that first that's causing this slightly less cool approach here. 🤷

jonallured:
Cool, this is green so merging.

PR #762, opened 2020-08-19
CSGN-307: Allow more submissions to be returned in queries

jonallured:
This PR cranks up the maximum number of submissions allowed to be returned all the way to 100. We have long been asking for more than 20 submissions using the `first` parameter to the query but it has not been supported because we have a default max of 20:

https://github.com/artsy/convection/blob/559a47b18a91189ce41c07f24edcc0a6054f1c0b/app/graphql/convection_schema.rb#L6

So then I looked up [the docs](https://graphql-ruby.org/pagination/using_connections.html#max-page-size) and read them more closely than I have before and allowing more submissions to be returned was as easy as passing the `max_page_size` option with a value of 100.

That's all this PR does and then I wrapped some tests around so that it was more clear how this property behaves at different values.

https://artsyproduct.atlassian.net/browse/CSGN-307

/cc @artsy/csgn-devs

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -61,7 +61,7 @@ module Types
 
     field :submissions,
           SubmissionConnectionType,
-          null: true, connection: true do
+          null: true, connection: true, max_page_size: 100 do
       description 'Filter all submission'
 
       argument :ids, [ID], required: false do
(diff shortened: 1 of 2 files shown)


app/graphql/types/query_type.rb
```diff
@@ -61,7 +61,7 @@ def submission(arguments)
 
     field :submissions,
           SubmissionConnectionType,
-          null: true, connection: true do
+          null: true, connection: true, max_page_size: 100 do
```
damassi, line 64: wow! this `max_page_size` prop is a sleeper

PR #773, opened 2020-08-26
Remove stale columns on offer table

jonallured:
This PR removes a couple columns from the offer table that were confusing and not used.

/cc @artsy/csgn-devs @louislecluse


jonallured:
Huh, looks like merge on green didn't work. 🤔

PR #775, opened 2020-08-26
CSGN-300: Set assigned to on submission page

jonallured:
This PR updates the submission detail screen so that an admin can assign the submission without having to go to the edit screen:

Screen Shot 2020-08-26 at 2 44 31 PM

https://artsyproduct.atlassian.net/browse/CSGN-300

/cc @artsy/csgn-devs

### DIFF
--- a/app/views/admin/submissions/show.html.erb
+++ b/app/views/admin/submissions/show.html.erb
@@ -60,7 +60,11 @@
             </div>
             <div class='bold-label single-padding-top'>Assigned To:</div>
             <div class='single-padding-top'>
-              <%= @submission.assigned_to || 'Unassigned' %>
+              <%= form_with model: [:admin, @submission] do |f| %>
+                <% options = [['Please select one', '']] + Convection.config.admin_names %>
+                <%= f.select :assigned_to, options %>
+                <%= f.submit 'Update' %>
+              <% end %>
             </div>
             <div class='single-padding-top'>
               <% if @submission.state == Submission::APPROVED %>
--- a/spec/system/submission_editing_spec.rb
+++ b/spec/system/submission_editing_spec.rb
@@ -46,18 +46,45 @@ describe 'Editing a submission', type: :feature do
   end
 
   context 'adding an admin to a submission' do
-    it 'displays that admin on the submission detail page' do
-      visit admin_submission_path(submission)
-      expect(page).to have_content('Unassigned')
+    context 'from the edit screen' do
+      it 'displays that admin on the submission detail page' do
+        visit admin_submission_path(submission)
+        expect(page).to_not have_select(
+                              'submission[assigned_to]',
+                              selected: 'Alice'
+                            )
 
-      click_on 'Edit'
-      expect(page).to have_content 'Assigned To:'
+        click_on 'Edit'
+        expect(page).to have_content 'Assigned To:'
 
-      select 'Alice', from: 'submission[assigned_to]'
-      click_button 'Save'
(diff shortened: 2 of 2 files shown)


app/views/admin/submissions/show.html.erb
```diff
@@ -60,7 +60,11 @@
             </div>
             <div class='bold-label single-padding-top'>Assigned To:</div>
             <div class='single-padding-top'>
-              <%= @submission.assigned_to || 'Unassigned' %>
+              <%= form_with model: [:admin, @submission] do |f| %>
```
pepopowitz, line 63: This is cool, I didn't know it would be so easy to add an "inline action" to the page. 

One thing I can't figure out, and it might be a really dumb question: what does `:admin` refer to in the `model`?

app/views/admin/submissions/show.html.erb
```diff
@@ -60,7 +60,11 @@
...
+                <% options = [['Please select one', '']] + Convection.config.admin_names %>
+                <%= f.select :assigned_to, options %>
+                <%= f.submit 'Update' %>
```
pepopowitz, line 66: How does rails choose the action/destination of this form submission?

app/views/admin/submissions/show.html.erb
jonallured: Yeah I wasn't expecting this to be easy but then I remembered that our pal @erikdstock added that notes section to this page with https://github.com/artsy/convection/pull/707 and so I looked at how he did it and it turned out I could copy him! His approach was to add a form to the page that posts to a Rails controller and then has that Rail controller redirect back to the submission show page. Since the submission has been updated prior to that second rendering, the page will be up-to-date.

This is not as cool as making a request with JS and then updating the DOM and having no page refresh, but that seemed like an acceptable tradeoff because this is an admin app.

Now to your question: that `form_with mode: [:admin, @submission]` stuff is a DSL for creating a form tag. The reason to send in an array of admin and submission is so that the form tag builder spits out a POST that has admin prefixed:

```ruby
# this route exists!
form_with model: [:admin, @submission] # => POST to /admin/submissions/:id
# this is a 404
form_with model: @submission # => POST to /submissions/:id
```

app/views/admin/submissions/show.html.erb
jonallured: See above!

app/views/admin/submissions/show.html.erb
pepopowitz: oh, I think I have a misconception about how it builds the form submission route. 

I was thinking it would take the model type (submission) and build a route from that, relative to where you currently were in the app. It sounds like it takes the model type and builds a route from that, relative to the root, unless you tell it otherwise. Do I have that right?

pepopowitz: "I could copy him!" - the easiest code ever written.

jonallured: Yeah it sounds like you're about there - if you want to read more, I would recommend this section:

https://guides.rubyonrails.org/form_helpers.html#binding-a-form-to-an-object

FWIW, I never get this stuff right on the first try and usually tinker with it and read docs until I have the magic incantation correct and then I pretty much never have to change it. 🤷

erikdstock: yes, i was surprised to learn about this as well... although if I remember correctly yuki pointed out there might be other, more semantic ways to do it but this served me well.

PR #780, opened 2020-08-28
CSGN-160: Add hold state to submission

jonallured:
This PR add a new state of a submission called `hold` that represents when a submission is temporarily not available. It looks like this:

Screen Shot 2020-08-28 at 10 12 00 AM

And then once set to hold, this can be altered like so:

Screen Shot 2020-08-28 at 10 11 48 AM

So that gives admins a lot of control about how they want to manage their submissions.

In terms of code, this was a pretty cheap change - I had previously extracted a class to represent the logic around which actions are appropriate for a given submission called `SubmissionStateActions` with unit tests and everything! So, I added a new item to the valid states on the Submission model and then came to that file and added the hold action. I updated a couple specs and this was done!

https://artsyproduct.atlassian.net/browse/CSGN-160

/cc @artsy/csgn-devs @louislecluse

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -19,6 +19,7 @@ class Submission < ApplicationRecord
     APPROVED = 'approved',
     PUBLISHED = 'published',
     REJECTED = 'rejected',
+    HOLD = 'hold',
     CLOSED = 'closed'
   ].freeze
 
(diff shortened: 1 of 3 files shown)


damassi:
@pepopowitz is OOO so gonna merge 👍

PR #783, opened 2020-08-31
Expose sold or bought in consignments via new field

ashleyjelks:
[Adds a new `consignments` field to view works that have either a  "sold" or "bought in " state.](https://artsyproduct.atlassian.net/browse/CSGN-333)

Screen Shot 2020-08-31 at 5 50 40 PM


spec/requests/api/graphql/queries/consignments_spec.rb
```diff
@@ -0,0 +1,278 @@
...
+      end
+
+      let(:query_inputs) do
+        "gravityPartnerId: \"#{
+          partner.gravity_partner_id
+        }\""
+      end
+
+      it 'returns only the requested consignments' do
+        post '/api/graphql', params: { query: query }, headers: headers
```
pepopowitz, line 167: **question [potentially blocking]:**

I don't think I totally understand this test. [The `context` name suggests we're looking for only "sent" consignments], and [the setup indicates that we want a consignment in the state of `consigned`](). The query specifies we are looking to limit results by `gravityPartnerId` though, not by "sent", though. And the test asserts that there are 2 results, but it doesn't tell me how those results are limited to "sent" results only.

Can you help me understand the goal of this test? There seems to be a mismatch between what it says it's going to do and what it does, or I have a misunderstanding about what it says it's going to do.


_Comment format based on [conventionalcomments.org](https://conventionalcomments.org/)_

app/graphql/types/consignment_type.rb
```diff
@@ -0,0 +1,19 @@
...
+      field :sale_date, String, null: true
+      field :sale_name, String, null: true
+      field :state, Types::ConsignmentStateType, null: true
+      field :id, ID, 'Uniq ID for this consignment', null: false
+      field :currency, String, null: true
+      field :internalID, ID, method: :id, null: true
+      field :sale_price_cents, Integer, null: true
+      field :submission, Types::SubmissionType, null: false
+    end
+  end
```
damassi, line 18: @ashleyjelks - are you running git commits with `--no-verify`? It looks like prettier isn't formatting code properly.

spec/requests/api/graphql/queries/consignments_spec.rb
ashleyjelks: Ah, these are leftover that should have been deleted! 👎  Just cleaned these up!

app/graphql/types/consignment_type.rb
ashleyjelks: I needed to configure it properly, should be good now.

spec/requests/api/graphql/queries/consignments_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+                    artistId
+                    category
+                    title
+                  }
+                }
+              }
+            }
+          }
+        GRAPHQL
+      end
```
jonallured, line 158: **suggestion [late-feedback]:** remove unused fields

I love this set of sorting specs, well done!! I just have a minor suggestion that we only ask for the fields in the query that will be used in the tests:

```ruby
      let(:query) do
        <<-GRAPHQL
        query {
          consignments(#{query_inputs}) {
              edges {
                node {
                  id
                }
              }
            }
          }
        GRAPHQL
      end
```

This way it's more clear to our future selves that these tests are strictly about sorting.

spec/requests/api/graphql/queries/consignments_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+        it 'returns the consignments sorted ascending by the id column' do
+          post '/api/graphql', params: { query: query }, headers: headers
+
+          expect(response.status).to eq 200
+          body = JSON.parse(response.body)
+
+          edges = body.dig('data', 'consignments', 'edges')
+          ids = edges.map { |edge| edge.dig('node', 'id') }
+
+          expect(ids).to eq %w[5 4 2 1]
```
jonallured, line 174: **praise:** nice specs!

Having just come off of a couple sprints worth of reading and writing sorting specs, I really appreciate how well this is done, thank you!! 🙏 

app/graphql/resolvers/consignments_resolver.rb
```diff
@@ -0,0 +1,45 @@
...
+    column_name = @arguments[:sort].keys.first
+    PartnerSubmission.column_names.exclude?(column_name)
+  end
+
+  def partner
+    Partner.find_by(gravity_partner_id: @arguments[:gravity_partner_id])
+  end
+
+  def conditions
+    { state: ['sold', 'bought in'] }
```
jonallured, line 36: **thought:** too restrictive?

It'll be interesting to see if/how this evolves. Will we end up wanting to fetch other consignment states? Not sure! 🤔 

Either way I love the discipline of only doing what was required to meet our current needs. Bravo!

jonallured (commented):
Hey @ashleyjelks sorry for the lateness, but I wanted to give you a little feedback. I don't expect any code changes and thanks for taking this on!!! ❤️ ❤️ ❤️

PR #795, opened 2020-09-14
Library Updates

jonallured:

```
$ bundle update
$ yarn upgrade --latest
```

```
$ ./bin/update
```


jonallured:
Note that this also includes a Rails bump for the security update they did last week.

PR #805, opened 2020-09-25
Configure Event Service to determine TLS flag from ENV.

artsyjian:
Under https://artsyproduct.atlassian.net/browse/PLATFORM-2834, we would like to switch Convection to use the  internal endpoint of RabbitMQ:

Staging: `rabbitmq.stg.artsy.systems`
Production: ` rabbitmq.prd.artsy.systems`

These endpoints, unlike the internet-facing ones currently in use, do not support TLS. Therefore, at the same time we make that switch, we will also have to tell Convection to not use TLS.

Convection is currently hardcoded to always use TLS, by `config.tls = true`. This PR is to let the flag be determined via ENV. After this PR, Convection continues to connect to RabbitMQ on its internet-facing endpoint and continues to use TLS.

Later on, we will switch Convection to Rabbit's internal endpoint, and it will be entirely done through ENV, by unsetting `RABBITMQ_CA_CERT,RABBITMQ_CLIENT_CERT,RABBITMQ_CLIENT_KEY,RABBITMQ_NO_VERIFY_PEER` and changing RABBITMQ_URL.

### DIFF
--- a/config/initializers/artsy_eventservice.rb
+++ b/config/initializers/artsy_eventservice.rb
@@ -4,7 +4,7 @@ Artsy::EventService.configure do |config|
   config.app_name = 'artsy-convection'
   config.event_stream_enabled = true
   config.rabbitmq_url = ENV['RABBITMQ_URL']
-  config.tls = true
+  config.tls = ENV['RABBITMQ_TLS']
   config.tls_ca_certificate = Base64.decode64(ENV['RABBITMQ_CA_CERT'] || '')
   config.tls_cert = Base64.decode64(ENV['RABBITMQ_CLIENT_CERT'] || '')
   config.tls_key = Base64.decode64(ENV['RABBITMQ_CLIENT_KEY'] || '')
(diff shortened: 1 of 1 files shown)


config/initializers/artsy_eventservice.rb
```diff
@@ -4,7 +4,7 @@
   config.app_name = 'artsy-convection'
   config.event_stream_enabled = true
   config.rabbitmq_url = ENV['RABBITMQ_URL']
-  config.tls = true
+  config.tls = ENV['RABBITMQ_TLS']
```
jonallured, line 7: **question:** config type change?

I'm not really familiar with `Artsy::EventService` so I could be off here, but note that this change will change the type being set for this config. What's here now is a `Boolean` and what ENV will return is always a `String` - does that matter?

izakp: Could this be set by conditionally checking for `RABBITMQ_CA_CERT` / `RABBITMQ_CLIENT_CERT` / `RABBITMQ_CLIENT_KEY`?

izakp (approved):
One minor comment but if we are going to drop TLS support after migrating it's inconsequential - so LGTM

config/initializers/artsy_eventservice.rb
artsyjian: Thanks @jonallured ! Noob mistake on my part. @izakp - Oh, that's indeed how Gravity does it, it sets `config.tls` boolean by presence of client cert or not. Okay. so I am copying Gravity's approach. Thanks.

PR #835, opened 2020-10-30
chore: switch from system tzdata to tzinfo-data gem for timezone happiness

pepopowitz:
Addresses [CX-744](https://artsyproduct.atlassian.net/browse/CX-744).

Like our other rails apps, this PR removes `tzdata` from the docker manifest, instead opting for the `tzinfo-data` gem to handle time zone data. 

See https://github.com/artsy/volt/commit/fbb846a9b95214b4b2d4cea3a6cca64a872b37da for an origin story of these changes, or something like https://github.com/artsy/diffusion/pull/347 for a similar PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,6 @@ RUN apk update && apk --no-cache --quiet add \
   nodejs \
   postgresql-dev \
   postgresql-client \
-  tzdata \
   yarn \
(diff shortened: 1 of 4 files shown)


starsirius (approved):
Thanks for helping with this issue! 🏆

PR #844, opened 2020-11-11
Feature: Adds OfferResponse model

sweir27:
Currently, there is no way for consignors to formally respond to offers. After receiving an offer via email, they can indicate their intent through a Google form, which Artsy admins must then copy over into convection's admin UI.

This PR is the first step toward allowing us to capture and display consignor's "offer responses" within convection.

**Intent of this PR**: Model out a basic `OfferResponse` to capture the data we want to collect from consignors.

**Out of scope**: API work, admin UI, etc. Those will come in subsequent PRs.

**Important bits of context**:
- Our admins do not want the consignor's response to actually change the state of the offer itself. Often the consignor has questions or would benefit from speaking with an Artsy admin before making a final call. This is the main reason to even have a separate `OfferResponse` data model!
- It doesn't seem bad to allow multiple responses to a single offer (and we don't intend to invest much in the UI-side, so this will likely be possible from clients). Therefore, an offer `has_many :offer_responses`. 

**Data model definition**:
```ruby
class AddOffersResponseModel < ActiveRecord::Migration[6.0]
  def change
    create_table :offer_responses do |t|
      ## Offer responses belong to offers, as consignors will be responding to offers directly.
      t.references :offer, foreign_key: true, index: true

      ## Right now, offer responses do not change the state of the offer, but in the future they could. This
      ## field captures the potential state of the offer after this response, but is not binding.
      t.string :intended_state, null: false

      ## In most cases, an Artsy admin will want to reach out to the consignor directly after receiving the response.
      t.string :phone_number

      ## This field is a catch-all for questions/comments/etc. that the consignor may have in response to the offer.
      t.text :comments

      ## This field matches the accepted rejection reasons on the Offer model so that we can easily "graduate" them.
      t.string :rejection_reason

      ## Standard!
      t.timestamps
    end
  end
end

```

### DIFF
--- a/app/models/offer_response.rb
+++ b/app/models/offer_response.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class OfferResponse < ApplicationRecord
+  INTENDED_STATES = [Offer::ACCEPTED, Offer::REJECTED, Offer::REVIEW].freeze
+
+  belongs_to :offer
+
+  validates :intended_state, inclusion: { in: INTENDED_STATES }
+  validates :rejection_reason,
+            inclusion: { in: Offer::REJECTION_REASONS }, allow_nil: true
+end
--- a/db/migrate/20201111212135_add_offers_response_model.rb
+++ b/db/migrate/20201111212135_add_offers_response_model.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddOffersResponseModel < ActiveRecord::Migration[6.0]
+  def change
+    create_table :offer_responses do |t|
+      t.references :offer, foreign_key: true, index: true
+      t.string :intended_state, null: false
+      t.string :phone_number
+      t.text :comments
+      t.string :rejection_reason
+      t.timestamps
+    end
+  end
+end
--- a/spec/models/offer_responses_spec.rb
+++ b/spec/models/offer_responses_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'support/gravity_helper'
+
+describe OfferResponse do
+  context 'intended_state' do
+    it 'allows only certain intended_states' do
+      expect(OfferResponse.new(intended_state: 'blah')).not_to be_valid
+      expect(OfferResponse.new(intended_state: Offer::ACCEPTED)).to be_valid
+      expect(OfferResponse.new(intended_state: Offer::SENT)).not_to be_valid
+    end
+
+    it 'is required' do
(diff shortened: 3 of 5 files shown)


spec/models/offer_responses_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    it 'is required' do
+      expect(OfferResponse.new()).not_to be_valid
+    end
+  end
+
+  context 'rejection_reason' do
+    it 'allows only certain rejection_reasons' do
+      expect(OfferResponse.new(intended_state: Offer::REJECTED, rejection_reason: 'Low estimate')).to be_valid
+      expect(OfferResponse.new(intended_state: Offer::REJECTED, rejection_reason: 'meow')).not_to be_valid
```
pepopowitz, line 22: 😻

pepopowitz (approved):
This makes sense to me!!!

It's red right now but it looks like it's a prettier-ruby issue (even though it says "tests" failed). I'm going to reassign to you to merge once you get that resolved.

app/models/offer_response.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class OfferResponse < ApplicationRecord
+  INTENDED_STATES = [
+    ACCEPTED = 'accepted',
+    REJECTED = 'rejected',
+    REVIEW = 'review'
+  ].freeze
```
joeyAghion, line 8: Can these refer to `Offer::STATES` rather than repeat a subset?

db/migrate/20201111212135_add_offers_response_model.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddOffersResponseModel < ActiveRecord::Migration[6.0]
+  def change
+    create_table :offer_responses do |t|
+      t.references :offer, foreign_key: true, index: true
+      t.string :intended_state, null: false
+      t.boolean :interested_in_offers
```
joeyAghion, line 8: Not sure I understand this. If it's meant to reflect whether the response is generally positive, can it just be `interested`? Or, can it be implicit based on whether a `rejection_reason` is specified?

sweir27: It's definitely confusing.

This field won't coexist with `rejection_reasons`.

It's meant to indicate that the consignor is _still interested in consigning with Artsy_, but not interested in taking the current offer. 

But maybe you're right that `interested` is a better way to capture that.

sweir27:
Addressed comments-- merging! Happy to revisit anything after-the-fact. 😄

PR #846, opened 2020-11-12
Feature: Adds APIs necessary for supporting consignor offer responses

sweir27:
The overarching goal of this is to pave the way for consignors to be able to respond to offers from a form on artsy.net. 

See previous modeling PR [here](https://github.com/artsy/convection/pull/844).

This PR:
1. Updates the `offer_resolver` to allow users to retrieve their own offers.
2. Adds a new mutation, `createConsignmentOfferResponse` to allow users to create responses to offers.

Screen Shot 2020-11-11 at 8 54 01 PM

I've left some more comments inline!


app/graphql/mutations/create_offer_response_mutation.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Mutations
+  class CreateOfferResponseMutation < Mutations::BaseMutation
+    argument :offer_id, ID, required: true
+    argument :intended_state, Types::IntendedStateType, required: true
+
+    argument :phone_number, String, required: false
+    argument :comments, String, required: false
+    argument :rejection_reason, String, required: false
```
sweir27, line 10: The input type here could be updated to match the database requirements. I worried that that would force us to keep multiple types up-to-date (and especially because the rejection reasons themselves are longer strings), so opted to rely on the ActiveRecord validation instead.

app/graphql/mutations/create_offer_response_mutation.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    field :consignment_offer_response, Types::OfferResponseType, null: true
```
sweir27, line 12: Externally, these types are prefaced with `consignment_` to disambiguate from BNMO offers (I'm guessing! I was also following convention here).

app/graphql/resolvers/create_offer_response_resolver.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class CreateOfferResponseResolver < BaseResolver
+  def valid?
+    return true if user?
```
sweir27, line 5: To keep it simple, this requires that a user ID is present in the JWT. Further authorization is done below.

app/graphql/resolvers/create_offer_response_resolver.rb
```diff
@@ -0,0 +1,38 @@
...
+    bad_argument_error =
+      GraphQL::ExecutionError.new("Can't access createConsignmentOfferResponse")
+    @error = bad_argument_error
+    false
+  end
+
+  def run
+    offer = Offer.find(offer_id)
+    matching_user =
+      offer.submission&.user&.gravity_user_id == @context[:current_user]
```
sweir27, line 16: Ensure that the offer they're responding to is indeed their own.

app/graphql/resolvers/offer_resolver.rb
```diff
@@ -9,15 +9,37 @@ def valid?
   end
 
   def run
-    partner.offers.find(@arguments[:id])
+    if partner? || admin? && partner.present?
```
sweir27, line 12: This basically says-- if the requesting JWT includes a `partner` (or `admin`) role, and you've passed in a `gravity_partner_id` argument, _then_ return just the partner's offer that matches the request. 

Otherwise we assume you're a user or admin looking for a single offer.

This means that the `gravity_partner_id` input is responsible for determining this scope.

app/graphql/types/query_type.rb
```diff
@@ -7,7 +7,7 @@ class QueryType < GraphQL::Schema::Object
 
       argument :id, ID, required: true
 
-      argument :gravity_partner_id, ID, required: true do
+      argument :gravity_partner_id, ID, required: false do
```
sweir27, line 10: Because we can now return offers for a user, this is no longer required.

spec/requests/api/graphql/queries/offer_spec.rb
```diff
@@ -65,7 +65,7 @@
         expect(offer_response).to eq nil
 
         error_message = body['errors'][0]['message']
-        expect(error_message).to eq "Can't access offer"
+        expect(error_message).to eq 'Offer not found'
```
sweir27, line 68: Changed the message here in order to not leak that an offer exists if you're not authorized to view it.

spec/requests/api/graphql/queries/offer_spec.rb
```diff
@@ -155,5 +200,52 @@
...
+      end
+
+      context 'as a valid user' do
+        let(:token) do
+          user_id = offer.submission.user.gravity_user_id
+          payload = { aud: 'gravity', sub: user_id, roles: 'user' }
+          JWT.encode(payload, Convection.config.jwt_secret)
+        end
+
+        it 'returns an error' do
```
sweir27, line 236: We don't allow users to access offers in the draft state. I _think_ this is okay, but should we restrict other states too?

spec/requests/api/graphql/mutations/create_consignment_offer_response_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+        consignmentOfferResponse {
+          id
+        }
+      }
+    }
+    GRAPHQL
+  end
+
+  describe 'invalid requests' do
+    context 'with an unauthorized request' do
```
pepopowitz, line 39: [non-blocking] Another useful test here would be `context 'with someone else's offer'` .. `it 'returns an error for that request'`. It _looks_ like that's accounted for in the underlying mutation, but it'd be nice to have a test to prove it. 

I would merge without this test but if you have the time & feel like it's easy to add before merging, it's a welcome addition.

spec/requests/api/graphql/queries/offer_spec.rb
pepopowitz: These are the states on an offer: 

```
  STATES = [
    DRAFT = 'draft',
    SENT = 'sent',
    ACCEPTED = 'accepted',
    REJECTED = 'rejected',
    LAPSED = 'lapsed',
    REVIEW = 'review'
  ].freeze
```

Aside from "draft", all of them seem like states that, if I were a consignor, I might want to reach out about if I'd changed my mind or had questions. I feel like we're good allowing responses in any states except "draft".

pepopowitz (approved):
❤️ ❤️ ❤️ 

This is amazing! 

I left a comment about one test that seems missing on creating an offer response. If you've got the time & it's easy to add, it'd be great! Let me know if you don't, and I'll merge this and add a test after.

app/graphql/mutations/create_offer_response_mutation.rb
jonallured: **question [non-blocking]:** minimum viable type checking

Can you elaborate on your worry about marking this as required? Seems like it would be better error feedback for one to pass an invalid string here than to leave it off and then get an error about the thing being required, no? I think what you're getting at is that you _could_ have created an enum with the possible valid values, but that seemed extra so you didn't. I guess my feeling is that marking this as a required string is a good middle ground. 🤷‍♂️ 

app/graphql/mutations/create_offer_response_mutation.rb
jonallured: Naming is hard! 😝

app/graphql/resolvers/create_offer_response_resolver.rb
jonallured: **question [non-blocking]:** check validity of intent sooner?

What happens if one tries to respond to an offer where they are not the gravity user? I wonder if this check belongs in the `valid?` method? 🤔 

jonallured:
Some random thoughts but this is super cool and I'm embarrassed to say that I did not know we were still using Google forms for the consignor's offer responses!! 😝

app/graphql/resolvers/create_offer_response_resolver.rb
sweir27: That check is just below!

Actually perhaps a question for you-- in the resolvers, I see a mix of approaches (some validations in `valid?` and others within the resolvers themselves).

I tried to keep the `valid?` methods fairly lean and based off of attributes that we get "for free" (i.e. whether you're an admin, user, partner, etc.) and to leave the more complex checks (for ownership, etc.) to the resolver method.

Many ways to approach this, curious if you have a strong opinion?

app/graphql/mutations/create_offer_response_mutation.rb
sweir27: Ah! The field is _not_ required, I just opted to not create an additional Enum type to capture the potential inputs given they'll also be checked at creation time.

But this is not a strong opinion, I can easily create an Enum to give quicker feedback (which in retrospect might be a huge help to the UI builder).

jonallured: OMG Jon read the code my dude. What you have here totally makes sense. I saw that validation check and assumed it was required. I see now that it's NOT required and we only validate that if present, it be a certain set of values.

In terms of building the UI, totally up to you if this enum helps. Could totally come in a follow up if you wanna keep moving.

app/graphql/resolvers/create_offer_response_resolver.rb
jonallured: Yeah I just read through the resolvers and you're right that there are various approaches. I think we are aligned on the line between `valid?` doing some sanity checking and the `run` method raising errors. Where I land on this is that `valid?` should probably stay out of the database. So we can check things like whether the token includes the admin flag or if the combination of query args are allowed, whatever. But checking Rails-level validations or doing something that hits the db is probably out of scope for the valid method.

So given that, I retract my question, haha. What you have here looks wonderful! 🙌

PR #849, opened 2020-11-13
Feature: Updates admin interface to handle offer responses

sweir27:
This PR updates the `offer`, `offers`, and `submission` views to support offer responses.

## Migration
Because we added a `counter_cache` on `Offer#offer_responses_count` in order to easily support filtering, we'll need to reset all counters.

```
Offer.find_each do |offer|
  Offer.reset_counters(offer.id, :offer_responses_count)
end
```

## Offer view
We show the "most recent offer response" (since technically there can be multiple) in the sidebar.
image

We show the entire list of offer responses at the bottom of the page.
image

## Submission view
We show the list of offers with the response obvious.

Note: We only show the `Response: ` text if the offer is in the `sent` state and has a response. Once it has "advanced" to a different state, we show that state instead.
image

## Offers view
Same as in the submission view, if the offer is in the `sent` state and has a response, we show the response. Otherwise we show the state.

image

We are also able to filter by offers that are `sent with response`. (i.e. in the state described above)
image

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_11_212135) do
+ActiveRecord::Schema.define(version: 2020_11_13_001511) do
   # These are extensions that must be enabled in order to support this database
   enable_extension 'pg_trgm'
   enable_extension 'plpgsql'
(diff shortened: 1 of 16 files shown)


db/schema.rb
```diff
@@ -87,6 +87,7 @@
     t.text 'partner_info'
     t.string 'deadline_to_consign'
     t.string 'sale_location'
+    t.integer 'offer_responses_count'
```
pepopowitz, line 90: [non-blocking] - for my education - this is associated with adding `counter_cache` to the model, right?

spec/views/admin/offers/show.html.erb_spec.rb
```diff
@@ -336,5 +340,85 @@
...
+          'Most recent offer response Interested in offer'
+        )
+      end
+
+      it 'shows multiple responses' do
+        Fabricate(
+          :offer_response,
+          offer: offer,
+          intended_state: Offer::REVIEW,
+          comments: 'Super interested but have questions.',
```
pepopowitz, line 407: 🦸‍♀️

pepopowitz (approved):
This looks good to me! I only had one question about a rails concept I've not seen before (`counter_cache`).

PR #852, opened 2020-11-13
feat: allow submission user to be edited

mdole:
Currently, when editing a submission, admins can search for & select a user, but after submission the user associated with the submission is not actually updated.

This PR adds that behavior, so that the user associated with a submission can be changed.

We also took the opportunity to clean up some naming :)

### DIFF
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -53,7 +53,7 @@ module Admin
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          params[:submission][:user_id]
+          submission_params[:user_id]
         )
       redirect_to admin_submission_path(@submission)
     rescue SubmissionService::SubmissionError => e
@@ -159,6 +159,7 @@ module Admin
         signature
         state
         title
+        user_id
         width
         year
       ]
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -5,8 +5,8 @@ class SubmissionService
   class SubmissionError < StandardError; end
 
   class << self
-    def create_submission(submission_params, current_user)
-      user = User.find_or_create_by!(gravity_user_id: current_user)
+    def create_submission(submission_params, gravity_user_id)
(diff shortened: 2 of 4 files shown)


app/controllers/admin/submissions_controller.rb
```diff
@@ -53,7 +53,7 @@ def create
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          params[:submission][:user_id]
+          submission_params[:user_id]
```
mdole, line 56: @jonallured after we chatted I figured we might as well clean this up too! tried it out locally and worked a-ok

mdole:
@sweir27 reassigning to you since Jon is out! Let me know if you have questions on any of these changes :)

app/views/admin/submissions/_form.html.erb
```diff
@@ -22,7 +22,7 @@
               </label>
               <div class='col-sm-10' id='user_selections_form'>
                 <%= f.text_field :user, class: 'form-control', id: 'user_search' %>
-                <%= f.hidden_field :user_id %>
+                <%= f.hidden_field :user_id, :value => @submission&.user&.gravity_user_id %>
```
mdole, line 25: Little context here: previously, when a user first loaded the form, the `value` in this field was the Convection user id, not the Gravity user id. After searching for a user, it would change to be the Gravity user id.

This was never a problem because we didn't permit the `user_id` param, so you could not update the user associated with a submission after it was created.

Because we're now permitting that param and we want to use it to update users associated with submissions, we needed to make this consistent! So now it's always the gravity ID of the user associated with the submission, and we find the convection user based on that ID.

app/services/submission_service.rb
```diff
@@ -16,7 +16,14 @@ def create_submission(submission_params, current_user)
     end
 
     def update_submission(submission, params, current_user: nil)
-      submission.assign_attributes(params)
+      if params[:user_id]
```
sweir27, line 19: Woohoo! This all seems fine to me. Would you mind adding a test to verify the behavior you've added here?

mdole: oop yep! good thought. added in 758b30d; that feel reasonable to you?

PR #860, opened 2020-11-20
feat: add source_artwork_id to submission schema and creation mutation

mdole:
@sweir27 made an excellent point about https://github.com/artsy/volt/pull/4745: we should start keeping _some_ kind of track of whether an artwork in Convection is associated with a Gravity artwork, especially as we do more and more syncing of existing artworks from partners to Convection.

This PR takes a step in that direction by adding a source ID field to submissions in Convection. After this is merged and deployed, I'll update the Volt PR to pass artwork IDs before we merge.

@sweir27 you think it's worth adding specs for CreateSubmissionMutation? ~~None exist at present, but it wouldn't be terribly hard to copy a different mutation spec and update the details.~~ I was wrong! The file was just named something sliiightly different from what I expected 😅 

Screenshot of me executing locally and creating a Convection submission with the id:
Screen Shot 2020-11-20 at 6 29 55 PM

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -439,6 +439,11 @@ input CreateSubmissionMutationInput {
   minimumPriceDollars: Int
   provenance: String
   signature: Boolean
+
+  """
+  If this artwork exists in Gravity, its ID
+  """
+  sourceArtworkID: String
   state: State
   title: String
   userAgent: String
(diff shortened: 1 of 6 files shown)


sweir27:
It might be worth adding a spec to confirm that: if a `sourceArtworkId` is passed in, it definitely gets saved to the db and returned in the graphQL response. Sometimes there are additional things to hook up to ensure that happens!

sweir27:
Also looks like there's a failure on CI-- let me know if you want some help sorting it out, the output on circle isn't very helpful at the moment.

mdole:
Thanks for the suggestion @sweir27! I found the existing spec and updated it, and looks like we're passing on CI 🙌

PR #868, opened 2020-12-04
Dependency updates

pepopowitz:
Partially generated with ./bin/update but ran into some weird schema-check issues that I don't think are an actual problem.


pepopowitz:
I think Circle issues prevented it from notifying GitHub that the checks had passed. The linked check is actually green, not pending. Merging!

PR #877, opened 2020-12-15
Do not merge: Re-write the Gravity GraphQL client

yuki24:
**This PR was sent for informational-purposes only. You may merge at our own discretion, but Yuki may not be able to remain responsible for shipping it all the way to production.**

There is [a transition note I have complied on notion](https://www.notion.so/artsy/Yuki-Transition-c6012e8ab1b5490e864e688da874ed22). In that note, there is the **Switching to Artemis** section. This PR demonstrates how it could be done in the actual code.


jonallured:
Thank you this is awesome!! ❤️

yuki24:
I didn't notice convection is open source now. Closing as this is more like an informational PR.

PR #881, opened 2020-12-16
Update consignments index page for CX-810

bhoggard:
Paired with @annacarey.

Completes https://artsyproduct.atlassian.net/browse/CX-810

Screen Shot 2020-12-15 at 2 57 09 PM


pepopowitz (approved):
That's a tight squeeze! Way to fit all those columns in there 💪

PR #883, opened 2020-12-16
fix: Don't pass null artist_ids to artist query

bhoggard:
If a submission is missing an artist, we'll see an error about fetching artist details if we pass in a null `artist_id`.


bhoggard:
Quick fix, so I'm going to go ahead and merge so I can deploy.

PR #907, opened 2021-01-06
fix:always return ActiveRecord relation from SubmissionMatch.find_all

bhoggard:
Props to @mikehrom for the SQL query.

### DIFF
--- a/app/models/submission_match.rb
+++ b/app/models/submission_match.rb
@@ -29,7 +29,7 @@ class SubmissionMatch
 
   def submissions_assigned_without_accepted_offer
     sql = <<SQL.squish
-    WITH wrap AS (
+    WITH submissions_with_counts AS (
       SELECT s.id as submission_id,
              COUNT(DISTINCT ps.id) AS count_partner_submissions,
              COUNT(DISTINCT ps.accepted_offer_id) AS count_accepted_offers
@@ -38,8 +38,8 @@ class SubmissionMatch
       GROUP BY 1
       )
       SELECT submission_id
-      from wrap
-      where count_partner_submissions = 0 OR count_accepted_offers = 0
+      FROM submissions_with_counts
+      WHERE count_partner_submissions = 0 OR count_accepted_offers = 0
 SQL
 
     ids = ActiveRecord::Base.connection.select_all(sql).rows.flatten
--- a/app/models/submission_match.rb
+++ b/app/models/submission_match.rb
@@ -13,19 +13,39 @@ class SubmissionMatch
 
   def find_all
     submissions =
-      Submission.not_deleted.where(query).includes(:user).order(order_by)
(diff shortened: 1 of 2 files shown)


app/models/submission_match.rb
```diff
@@ -13,20 +13,40 @@ def initialize(params)
...
+      SELECT s.id as submission_id,
+             COUNT(DISTINCT ps.id) AS count_partner_submissions,
+             COUNT(DISTINCT ps.accepted_offer_id) AS count_accepted_offers
+      FROM submissions s
+       LEFT JOIN partner_submissions ps on ps.submission_id = s.id
+      GROUP BY 1
+      )
+      SELECT submission_id
+      from wrap
+      where count_partner_submissions = 0 OR count_accepted_offers = 0
```
pepopowitz, line 42:
```suggestion
    WITH wrap AS (
      SELECT s.id as submission_id,
             COUNT(DISTINCT ps.id) AS count_partner_submissions,
             COUNT(DISTINCT ps.accepted_offer_id) AS count_accepted_offers
      FROM submissions s
       LEFT JOIN partner_submissions ps on ps.submission_id = s.id
      GROUP BY 1
      )
      SELECT submission_id
      FROM wrap
      WHERE count_partner_submissions = 0 OR count_accepted_offers = 0
```

Non-blocking & not a big deal. I'd imagine we don't have any rubocopping on format of SQL queries, so I think we're on our own to manage these kinds of consistencies.

pepopowitz, line 42:
```suggestion
    WITH submissions_with_counts AS (
      SELECT s.id as submission_id,
             COUNT(DISTINCT ps.id) AS count_partner_submissions,
             COUNT(DISTINCT ps.accepted_offer_id) AS count_accepted_offers
      FROM submissions s
       LEFT JOIN partner_submissions ps on ps.submission_id = s.id
      GROUP BY 1
      )
      SELECT submission_id
      from submissions_with_counts
      where count_partner_submissions = 0 OR count_accepted_offers = 0
```

Also non-blocking & not a big deal, but I personally find it helpful for understanding to give contextual names to common table expressions.

pepopowitz (commented):
The changes are clear & look good to me, but I don't really understand why they're needed. I would have expected the old code to also be returning an ActiveRecord relation. Can you give more context or link me to some ActiveRecord docs that would make it more clear to me why the old version wasn't working?

app/models/submission_match.rb
mikehrom: +1 to the contextual name 👍

bhoggard:
The old code, by dropping out of `ActiveRecord` code and using methods from `Enumerable`, converted the relation to a regular array.

bhoggard:
#mergeongreen

bhoggard:
Merge on green didn't work. I'm going to merge this and test on staging.

PR #923, opened 2021-01-21
fix: Use conventional structure for middleware file

jonallured:
We noticed that staging was not deploying correctly and did notice an this exception:

```
expected file /Users/jon/code/convection/lib/middleware/jwt_middleware.rb to define constant Middleware::JwtMiddleware, but didn't (Zeitwerk::NameError
```

After some tinkering I was able to reproduce locally like this:

```
$ RAILS_ENV=production ./bin/server
```

And now that I'm typing this out that error message is actually pretty good! It's telling us that Zeitwork is expecting one thing but we're doing another. Our choices were to:

* do what Zeitwork is expecting and add a `Middleware` namespace to that file
* move that file out of the `middleware` folder so that Zeitwork wouldn't expect a namespace at all
* add `lib/middleware` to the list of "roots" that Zeitwork knows about
* revert back to the `classic` mode of Zeitwork

I decided to just move that damn file. 🤷 

Lots more chatter here: https://artsy.slack.com/archives/CA8SANW3W/p1611231374001500


artsyjian (approved):
Nice one!

pepopowitz (approved):
THANK YOU @jonallured !!!!!!!

izakp:
Thanks @jonallured !

PR #939, opened 2021-02-03
Update dependencies

pepopowitz:
Started with bin/update, but it failed halfway through due to some harmless schema updates.


Gemfile
```diff
@@ -23,7 +23,7 @@ gem 'ddtrace'
 gem 'decent_exposure' # for safely referencing variables in views
 gem 'gemini_upload-rails' # for admins to upload images
 gem 'graphiql-rails' # A lovely interface to the API
-gem 'graphql'
+gem 'graphql', '~>1.11.6'
```
pepopowitz, line 26: @jonallured did I do this right? From what I could understand from docs, this means "any patch version of 1.11 at 1.11.6 or higher", which is what I want, because 1.12 breaks a handful of tests.

jonallured: I think so! FWIW, I tend to pin straight to the current version like so:

```
gem 'graphql', '1.11.6'
```

And then in the commit message I'll say why so there's a paper trail. In my experience it's rare that we'll want to keep the pin and pick up patch versions. Plus I can never confidently remember this syntax and the pins are meant to be temporary. 🤷

pepopowitz: Cool, thanks -- I'm going to leave it as `~>1.11.6` for now to avoid having to re-bundle-update (it pulled in 1.11.7), but I'll keep this in miind for next time.

PR #945, opened 2021-02-05
chore: pin omniauth to <2 due to breaking change in login

pepopowitz:
v2 of omniauth includes changes that break our login flow. This PR pins the dependency to `<2`. 

Issue was originally found [in volt](https://github.com/artsy/volt/pull/4893), thanks to @starsirius for bringing it to our attention that it login was broken here too.


pepopowitz:
Gonna merge on green to see how it does in staging

PR #975, opened 2021-03-05
feature:Add consignments_manager to allowed roles

bhoggard:
Ticket: https://artsyproduct.atlassian.net/browse/CX-1142

I haven't received a list of consignments admins yet, so I'm temporarily allowing both admins and consignments managers to access the application.

I verified that this works with a user on staging.


jonallured (approved):
LGTM!! 🎉

PR #1022, opened 2021-04-09
[CX-633]: feature(submission): show artist supply priority on #show

amy-mkrtchyan:
- show artist supply priority on #show
- remove artists Gravity call from submission

Refs [CX-633]

P1_artist
P2_artist
unknown_artist

[CX-633]: https://artsyproduct.atlassian.net/browse/CX-633

### DIFF
--- a/app/controllers/concerns/graphql_helper.rb
+++ b/app/controllers/concerns/graphql_helper.rb
@@ -7,7 +7,7 @@ module GraphqlHelper
     <<~GQL
       query artistsDetails($ids: [ID!]!){
         artists(ids: $ids){
-          #{fields.join(', ')}
+          #{[:id, *fields].join(', ')}
         }
       }
     GQL
@@ -26,7 +26,7 @@ module GraphqlHelper
   def artists_names_query(artist_ids)
     artist_details_response =
       Gravql::Schema.execute(
-        query: artist_query_builder(fields: [:id, :name]),
+        query: artist_query_builder(fields: [:name]),
         variables: { ids: artist_ids.compact.uniq }
       )
     if artist_details_response[:errors].present?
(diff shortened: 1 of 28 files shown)


app/controllers/concerns/graphql_helper.rb
```diff
@@ -3,15 +3,15 @@
...
-  ARTISTS_DETAILS_QUERY =
-    '
-  query artistsDetails($ids: [ID!]!){
-    artists(ids: $ids){
-      id
-      name
-    }
-  }
-  '
+  def artist_query_builder(fields: [])
```
bhoggard, line 6: Nice refactor 👍

app/models/submission.rb
```diff
@@ -133,12 +135,6 @@ def reviewed_by_user
     nil
   end
 
-  def artist
```
bhoggard, line 136: I don't think we can remove this, otherwise the submission edit page's artist field will appear blank when and admin is editing it.

spec/helpers/submissions_helper_spec.rb
```diff
@@ -206,4 +206,21 @@
...
   end
+
+  context 'artist_supply_priority' do
+    context 'when is_p1' do
+      subject { helper.artist_supply_priority(is_p1: true) }
+      it { is_expected.to eq 'P1' }
+    end
+
+    context 'when target_supply' do
+      subject { helper.artist_supply_priority(is_p1: true) }
```
bhoggard, line 217: This should be a `target_supply` / `p2` test I think?

spec/models/submission_spec.rb
```diff
@@ -189,31 +189,6 @@
     end
   end
 
-  context 'artist' do
```
bhoggard, line 192: If we add back the gravity query, this test should come back too.

bhoggard (changes requested):
I think there need to be a couple of changes before merging.

app/models/submission.rb
amy-mkrtchyan: You are right. Moved Gravql call to submissions_controller `before_action` for `edit` and `show`. Now it will appear in edit page too.

spec/helpers/submissions_helper_spec.rb
amy-mkrtchyan: Yes :) Fixed.

PR #1026, opened 2021-04-10
improvement: Add extra specs for rejected offer, remove attr accessor

amy-mkrtchyan:
Refs CX-614, CX-633


bhoggard (approved):
Thanks for the follow-up.

PR #1030, opened 2021-04-12
fix: bump pg-data-sync version due to rds at 12.5 now.

artsyjian:
use pg-data-sync:12 image which is based on postgres 12 which matches rds version.


jonallured:
Oops, sorry for not merging sooner!! 🙌

PR #1033, opened 2021-04-14
don't blow up on empty string for artist_id

bhoggard:
Somehow a submission ended up with an `artist_id` that was a blank string, which caused the whole query to blow up.

In a rails console:

```ruby
artist_ids = ["1", "", nil, "2", "3", "3"]

artist_ids.compact.uniq # => ["1", "", "2", "3"]

artist_ids.uniq.select(&:present?) # => ["1", "2", "3"]


pepopowitz (approved):
💪

PR #1042, opened 2021-04-21
fix: removed mistaken named parameter in call to artists_names_query

bhoggard:
I missed this in code review, which meant all of the sidekiq jobs to send out partner digest emails have been failing since April 9.

https://artsyproduct.atlassian.net/browse/CX-1350


amy-mkrtchyan:
My fault, sorry!

bhoggard:
We should probably have tests for this, or I should have caught it. No blame.

amy-mkrtchyan:
Agree.

PR #1066, opened 2021-05-12
feat: add invoice number to partner consignments

leamotta:
Ticket: https://artsyproduct.atlassian.net/browse/CX-1392

Add partner section and invoice number in edit page of the partner consignments.
Decided to only show the invoice number field if the "Partner invoiced?" field is checked. The flow can be seen below:

https://user-images.githubusercontent.com/7518671/117990803-cbe39a00-b33d-11eb-8f43-29687ad4dab8.mov

Note: If the admin decides to uncheck the "Partner invoiced?" field, the invoice number will not be erased from the database, and it will not be visible from the page until "Partner invoiced?" is checked again.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -189,6 +189,16 @@ enum ConsignmentSort {
   ID_DESC
 
   """
+  sort by invoice_number in ascending order
+  """
+  INVOICE_NUMBER_ASC
+
+  """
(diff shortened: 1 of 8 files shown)


bhoggard:
OK, they are strings.

bhoggard (approved):

PR #1075, opened 2021-05-25
 feat(submission): Add UTM fields

angelagasparyan:
Refs [CX-1399]

[CX-1399]: https://artsyproduct.atlassian.net/browse/CX-1399


olerichter00 (approved):
Looks great 👍

bhoggard (changes requested):
Based on the description in https://artsyproduct.atlassian.net/browse/CX-1269 we need to add these to display in

https://github.com/artsy/convection/blob/master/app/views/admin/submissions/show.html.erb

if present.

angelagasparyan:
> Based on the description in https://artsyproduct.atlassian.net/browse/CX-1269 we need to add these to display in
> 
> https://github.com/artsy/convection/blob/master/app/views/admin/submissions/show.html.erb
> 
> if present.

This is only first subtask

bhoggard:
OK, I see. We should be OK once the conflict is fixed.

angelagasparyan:
@bhoggard conflict is fixed ))

bhoggard (approved):

PR #1082, opened 2021-05-28
feat: Added rake tasks to lapse all offers sent or reviewed before a certain date

Gon4ar5:
Resolves: [CX-1509](https://artsyproduct.atlassian.net/browse/CX-1509) and [CX-1511](https://artsyproduct.atlassian.net/browse/CX-1511)

I have added two rake tasks for lapsing offers that are in the status of sent or reviewed before a certain date. 
As an example of launching: `rake offers:lapse_review ["2021-09-01"]`

### DIFF
--- a/lib/tasks/offers.rake
+++ b/lib/tasks/offers.rake
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+namespace :offers do
+  desc 'Lapse all offers that were sent before a certain date'
+  task :lapse_sent, [:date] => :environment do |_task, args|
+    raise 'Please supply a date until which sent offers should be considered to lapse!' unless args[:date]
+
+    datetime = args[:date].to_datetime
+    Offer.where("state = ? AND (sent_at < ? OR sent_at IS NULL)", "sent", datetime).update_all(state: "lapsed")
+  end
+
+  desc 'Lapse all offers that were reviewed before a certain date'
+  task :lapse_review, [:date] => :environment do |_task, args|
+    raise 'Please supply a date until which reviewed offers should be considered to lapse!' unless args[:date]
+
+    datetime = args[:date].to_datetime
+    Offer.where("state = ? AND review_started_at < ?", "review", datetime).update_all(state: "lapsed")
+  end
+end
--- a/lib/tasks/offers.rake
+++ b/lib/tasks/offers.rake
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# rubocop:disable Lint/RedundantCopDisableDirective
+# rubocop:disable Rails/SkipsModelValidations
+
 namespace :offers do
   desc 'Lapse all offers that were sent before a certain date'
   task :lapse_sent, [:date] => :environment do |_task, args|
@@ -17,3 +20,6 @@ namespace :offers do
     Offer.where("state = ? AND review_started_at < ?", "review", datetime).update_all(state: "lapsed")
   end
 end
+
+# rubocop:enable Rails/SkipsModelValidations
+# rubocop:enable Lint/RedundantCopDisableDirective
--- a/spec/tasks/offers_spec.rb
+++ b/spec/tasks/offers_spec.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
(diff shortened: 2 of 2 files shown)


lib/tasks/offers.rake
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+namespace :offers do
+  desc 'Lapse all offers that were sent before a certain date'
+  task :lapse_sent, [:date] => :environment do |_task, args|
+    raise 'Please supply a date until which sent offers should be considered to lapse!' unless args[:date]
+
+    datetime = args[:date].to_datetime
+    Offer.where("state = ? AND (sent_at < ? OR sent_at IS NULL)", "sent", datetime).update_all(state: "lapsed")
```
olerichter00, line 9: Actually, Rubocop gives a warning about `update_all` skipping validations. It might be worth it to look at ActiveRecord Batch operations (https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Batches.html) and do something like:

```
Offer.where("state = ? AND (sent_at < ? OR sent_at IS NULL)", "sent", datetime).in_batches do |offer|
  relation.update(state: "lapsed")
end
```

lib/tasks/offers.rake
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  desc 'Lapse all offers that were reviewed before a certain date'
+  task :lapse_review, [:date] => :environment do |_task, args|
+    raise 'Please supply a date until which reviewed offers should be considered to lapse!' unless args[:date]
+
+    datetime = args[:date].to_datetime
+    Offer.where("state = ? AND review_started_at < ?", "review", datetime).update_all(state: "lapsed")
```
olerichter00, line 17: Same here (see https://app.circleci.com/pipelines/github/artsy/convection/2672/workflows/cb117bc9-7f10-4131-a889-d2af654de8db/jobs/3274)

olerichter00 (changes requested):
Looks good so far! 

1. We have to address the Rubocop warning.
2. I'd add some specs just to be sure not to accidentally change the wrong offers.

bhoggard:
I think it's OK to use `update_all!` (with the `!`) and disable the `Rails/SkipsModelValidations` cop in this file. @olerichter00 I think these are pretty simple, but it we're going to add tests, how would you recommend doing that with a rake task? I normally would move the work into a service object for testing in this case, but I don't feel strongly about adding tests for this.

Gon4ar5:
> I think it's OK to use `update_all!` (with the `!`) and disable the `Rails/SkipsModelValidations` cop in this file. @olerichter00 I think these are pretty simple, but it we're going to add tests, how would you recommend doing that with a rake task? I normally would move the work into a service object for testing in this case, but I don't feel strongly about adding tests for this.



> Yeah, `update_all!` should be fine as well here. Maybe it's just me being paranoid about messing around with prod data :)
> 
> Testing rake tasks is pretty similar to testing a service. Something like this should work:
> 
> ```
> # frozen_string_literal: true
> 
> require "rails_helper"
> 
> Rails.application.load_tasks
> 
> describe "Offers lapse sent rake task" do
>   subject(:invoke_task) do
>     Rake.application.invoke_task("offers:lapse_sent[#{date}]")
>   end
> 
>   let(:date) { "2021-09-01" }
>   let(:past_offer) { Fabricate(:offer, state: "sent", sent_at: date.to_date - 1.day) }
>   let(:future_offer) { Fabricate(:offer, state: "sent", sent_at: date.to_date + 1.day) }
>   # ... add other offers that shouldn't change
> 
>   it "testing..." do
>     expect { invoke_task }.to change { past_offer.reload.state }.from("sent").to("lapsed")
>     expect { invoke_task }.not_to change { future_offer.reload.state }
>     # test other offers
>   end
> end
> ```

Maybe I don’t know something, but such a method exists?
Screenshot 2021-05-31 at 12 54 36
@olerichter00 Thanks for the example test

olerichter00:
Hmm, I don't see `update_all!` as well. Then just let's use `update_all`

olerichter00:
Looks great! Let's ship it 🚢

olerichter00 (approved):

PR #1091, opened 2021-06-22
update: add coa fields to submissions table and mutation

kajatiger:
Resolving this ticket: [PX-4130](https://artsyproduct.atlassian.net/browse/PX-4130)

Paired with @mdole on this.

Connected to the `create_consignment_submission` action over on the [volt artworks controller](https://github.com/artsy/volt/blob/5d4da239203b1c13e7e9151efbac0c0cd03375ae/app/controllers/artworks_controller.rb#L238), where we were introducing a bug by not actually having the fields for the params that we send over to convection. 

This is also confirmed as useful by Sarah Shelburn 👍 
 
In order to fix this bug, we need to first introduce it to the schema here in convection. 
Follow up tasks:
- [ ] copy the new schema to metaphysics and maybe stitch
- [x] implement both new fields into the [view on convection](https://github.com/artsy/convection/blob/835432291c8c1fc56a1179e2bc42740d92f90dfa/app/views/admin/submissions/_submission_details.html.erb)

Screenshot of where to add the view:
Bildschirmfoto 2021-06-22 um 13 16 30

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_06_02_193423) do
+ActiveRecord::Schema.define(version: 2021_06_22_110240) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_trgm"
@@ -188,6 +188,8 @@ ActiveRecord::Schema.define(version: 2021_06_02_193423) do
     t.string "utm_source"
     t.string "utm_medium"
     t.string "utm_term"
+    t.boolean "coa_by_authenticating_body"
+    t.boolean "coa_by_gallery"
     t.index ["consigned_partner_submission_id"], name: "index_submissions_on_consigned_partner_submission_id"
     t.index ["ext_user_id"], name: "index_submissions_on_ext_user_id"
     t.index ["primary_image_id"], name: "index_submissions_on_primary_image_id"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -188,6 +188,8 @@
     t.string "utm_source"
     t.string "utm_medium"
     t.string "utm_term"
+    t.string "coa_by_authenticating_body"
+    t.string "coa_by_gallery"
```
mdole, line 192: To match with the [gravity definition](https://github.com/artsy/gravity/blob/50aeb677c8cddc401161b65bffa09dba57167f82/app/models/util/artwork_condition.rb), these should actually be booleans both here and in the GraphQL type :)

mdole (changes requested):
Woohoo, thanks for taking this on @kajatiger! One change request: the types for both COA fields should be `boolean`, since it's just a "yes this work has a certificate of this type" field (and because that's how it is in Gravity).

I also think it would be OK to update Convection's UI in this same PR if you want to, but a separate PR is fine too :)

db/schema.rb
kajatiger: Oh thanks! That's even easier. Should there be a default value like `false` if none is given? Or is `nil` okay..?

mdole: I think `nil` is better because it more clearly indicates that we don't know whether or not a work has a certain type of COA - `false` implies that we're certain it _doesn't_ have that type of COA.

mdole (approved):

jonallured (approved):
LGTM! 🎉

PR #1098, opened 2021-06-29
feat[CX-1495] Admin can assign a cataloguer as secondary assignee

Gon4ar5:
Resolves [CX-1495](https://artsyproduct.atlassian.net/browse/CX-1495)

Screenshot 2021-06-29 at 13 40 25 Screenshot 2021-06-29 at 13 40 35
Screenshot 2021-06-29 at 13 39 54
Screenshot 2021-06-29 at 13 39 51
Screenshot 2021-06-30 at 15 43 19


olerichter00 (approved):
LGTM

bhoggard (changes requested):
Rather than hard-wire the names, can we use the same approach we're using for admins, with an associated user ID? https://github.com/artsy/convection/blob/master/config/initializers/global_constants.rb

Gon4ar5:
> Rather than hard-wire the names, can we use the same approach we're using for admins, with an associated user ID? https://github.com/artsy/convection/blob/master/config/initializers/global_constants.rb

Good, but how can I find their user ID?

bhoggard:
You can look them up as you did [here](https://github.com/artsy/convection/pull/1089). If looking them up by name doesn't work, you can ask for the emails, or find them in their Slack profiles.

bhoggard (approved):

PR #1101, opened 2021-07-01
feat: Add category filter on submission page

Gon4ar5:
Resolves [CX-1393](https://artsyproduct.atlassian.net/browse/CX-1393)
Screenshot 2021-07-01 at 12 58 32


olerichter00 (approved):
Good job - it looks very good!

PR #1104, opened 2021-07-05
feat(CX-654): Adding a Notes Field that is linked to user

Gon4ar5:
Resolves [CX-654](https://artsyproduct.atlassian.net/browse/CX-654)

Screenshot 2021-07-06 at 04 13 14


db/migrate/20210702113010_assign_note_with_partner.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AssignNoteWithPartner < ActiveRecord::Migration[6.1]
+  def change
+    add_column :notes, :assign_with_partner, :boolean
```
bhoggard, line 5: Why is this called `assign_with_partner` when it's about adding a note for a user?

olerichter00 (commented):
I'm wondering if there is a specific reason why we have to add the flag `assign_with_partner` to `notes`. I would rather expect `notes` to have a one-to-many relationship with `users` (the consignor). 

Designing the relationship like this

```
class Note
  ...
  belongs_to :user
  ...
end

class User
  ...
  has_many :notes, dependent: :nullify
  ...
end
```

would give us the possibility to fetch all notes of a user/consignor with `@user.notes` or `@submission.user.notes` which IMHO would be easier to read.

Do you think that would be feasible?

app/views/admin/notes/_form.html.erb
```diff
@@ -1,5 +1,9 @@
 <%= form_with model: [:admin, note], local: true do |f| %>
   <%= f.hidden_field :submission_id %>
   <%= f.text_area :body, required: "required" %>
-  <%= f.submit 'Create' %>
-<% end %>
\ No newline at end of file
+  <%= f.label :add_note_to_user do %>
+    <%= f.check_box :add_note_to_user %>
+    Add a note to the user
```
olerichter00, line 6: @bhoggard what do you think would be a descriptive description here? Do you think something like this is enough to understand what this checkbox means?

app/controllers/admin/submissions_controller.rb
```diff
@@ -79,7 +79,7 @@ def show
       @partner_submissions_count =
         notified_partner_submissions.group_by_day.count
       @offers = @submission.offers
-      @notes = @submission.notes
+      @notes = @submission.notes + @submission.user.notes
```
olerichter00, line 82: I wonder if we should visually show if the note is about the submission or the user. I can imagine that it could look confusing to just mix all notes. Do you have a good idea how we could do that?

olerichter00 (commented):
Looks very good - thanks a lot for the changes!

app/controllers/admin/submissions_controller.rb
olerichter00: We could also ask whoever requested this feature.

bhoggard: @olerichter00 you could ask in the ticket. They're pretty responsive.

Gon4ar5:
@olerichter00 @bhoggard 
Do you think this will be enough to visually distinguish these notes?
Screenshot 2021-07-13 at 12 22 06

olerichter00:
That looks pretty good to me!

olerichter00 (approved):

PR #1105, opened 2021-07-12
feat: Introduce admin users

olerichter00:
Addresses [CX-1647]

### Description

- Adds admin users to replace the `ADMIN` and `CATALOGUERS` consts.
- Adds admin users interface for CRUD actions.

Screenshot 2021-07-12 at 19 27 40

### Follow Ups

- Run `bundle exec rake admin_users:create`
- Remove `config/initializers/global_constants.rb`

[CX-1647]: https://artsyproduct.atlassian.net/browse/CX-1647

### DIFF
--- a/app/views/admin/admin_users/_form.html.erb
+++ b/app/views/admin/admin_users/_form.html.erb
@@ -1,0 +1,45 @@
+<div class='container'>
+  <%= form_for @admin_user, url: @admin_user.persisted? ? admin_admin_user_path(@admin_user) : admin_admin_users_path do |f| %>
+  <% if @admin_user.errors.any? %>
+  <div id="error_explanation">
+    <h2>
+      <%= "#{pluralize(@admin_user.errors.count, "error")} prohibited this admin from being saved:" %>
+    </h2>
+    <ul>
+      <% @admin_user.errors.full_messages.each do |message| %>
+      <li>
+        <%= message %>
+      </li>
+      <% end %>
+    </ul>
+  </div>
+  <% end %>
+  <div class="field">
+    <%= f.label :name %>
+    <%= f.text_field :name %>
+  </div>
+  <div class="field">
+    <%= f.label :gravity_user_id %>
+    <%= f.text_field :gravity_user_id %>
+  </div>
+  <div class="field">
+    <%= f.label :assignee %>
+    <%= f.check_box :admin %>
+  </div>
+  <div class="field">
+    <%= f.label :cataloguer %>
+    <%= f.check_box :cataloguer %>
+  </div>
+  <div class="field">
+    <%= f.label :super_admin %>
+    <%= f.check_box :super_admin %>
+  </div>
+  <div class="actions">
+    <div class='row triple-padding-top'>
+      <div class='col-md-12'>
+        <%= f.submit 'Save', class: 'btn btn-primary btn-full-width' %>
+      </div>
+    </div>
+  </div>
+  <% end %>
+</div>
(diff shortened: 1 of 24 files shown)


app/views/admin/admin_users/_form.html.erb
```diff
@@ -0,0 +1,46 @@
+<div class='container'>
+
```
olerichter00, line 2:
```suggestion
```

Gon4ar5:
https://github.com/artsy/convection/blob/olerichter00/CX-1647/add-admin-model/app/views/admin/submissions/show.html.erb#L115

I think it needs to be changed from
`<% options = [['Please select one']] + CATALOGUERS.collect { |grav_id, name| [name, grav_id] } %>`
to
`<% options = [['Please select one']] + AdminUser.cataloguers.map { |cataloguer| [cataloguer.name, cataloguer.gravity_user_id] } %>`

Gon4ar5:
> https://github.com/artsy/convection/blob/olerichter00/CX-1647/add-admin-model/app/views/admin/submissions/show.html.erb#L115
> 
> I think it needs to be changed from
> `<% options = [['Please select one']] + CATALOGUERS.collect { |grav_id, name| [name, grav_id] } %>`
> to
> `<% options = [['Please select one']] + AdminUser.cataloguers.map { |cataloguer| [cataloguer.name, cataloguer.gravity_user_id] } %>`

Is it possible to leave comments on unchanged code on github? I do not understand😅

olerichter00:
Good catch, @Gon4ar5 ! I forgot to update the cataloguers. I updated all appearances  of `CATALOGUERS`

Gon4ar5:
And also (purely from my visual point of view😁 ) it would be better to make the buttons of the same size and "align: center" on some elements, something like this
Screenshot 2021-07-15 at 17 09 44

Gon4ar5 (approved):
Very cool! Thank you for what you did🔥 🔥 🔥

olerichter00:
> And also (purely from my visual point of view😁 ) it would be better to make the buttons of the same size and "align: center" on some elements, something like this
> Screenshot 2021-07-15 at 17 09 44

@Gon4ar5 Yes, that looks way better. Can you push these changes? I just kept it as simple and didn't pay too much attention to how things look...

app/views/admin/admin_users/index.html.erb
```diff
@@ -11,41 +11,41 @@
...
       <%= admin_user.cataloguer ? "Yes" : "No" %>
     </div>
-    <div class='col-md-1'>
+    <div class='col-md-1 align-center'>
       <%= admin_user.super_admin ? "Yes" : "No" %>
     </div>
-    <div class='col-md-1'>
-      <%= button_to 'Edit', edit_admin_admin_user_path(admin_user), method: :get, class: 'btn btn-small btn-primary pull-right' %>
+    <div class='col-md-2'%>
+      <%= button_to '&emsp;Edit&emsp;'.html_safe, edit_admin_admin_user_path(admin_user), method: :get, class: 'btn btn-small btn-primary pull-right', style: 'float: center;'%>
```
Gon4ar5, line 45: Not really sure if this approach is correct, but I haven't come up with anything better🙈

Gon4ar5 (approved):

app/views/admin/admin_users/index.html.erb
olerichter00: 👍

lib/tasks/offers_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
joeyAghion, line 3: I think this is causing staging deploys [to fail recently](https://app.circleci.com/pipelines/github/artsy/convection?branch=master). It should go under `spec/` if it's a test, so the deployed environments don't try to load it.

PR #1107, opened 2021-07-15
(CX-1513) Automatic emails sent to partners come from the email of assignee

Gon4ar5:
Resolves [CX-1513](https://artsyproduct.atlassian.net/browse/CX-1513)

Description:
Emails to partners are sent from our personal emails so that if the partner replies directly to an offer/rejection email, the reply is sent back to our personal inbox. This would avoid responses to get lost in the consign@ inbox which can fall through the cracks.


bhoggard (approved):

olerichter00 (approved):
👍

PR #1108, opened 2021-07-21
feat(CX-1711): Add saved state to offers

Gon4ar5:
Resolve [CX-1711](https://artsyproduct.atlassian.net/browse/CX-1711)

Screenshot 2021-07-21 at 14 11 04

After save you can send
Screenshot 2021-07-21 at 14 11 17

What is the problem?
Convection was built to flow to Auction House (AH) digest, and to approve organic pipeline submissions. We are now predominantly using it to approve and track ALL consignments, so all non-organic consignments have to go through the same offer approval process in the system. This includes an automatic email sent to the consignor letting them know they have a new offer on their submission - a very bad and confusing consignor experience when they have already negotiated with specialists and agreed on sale placement and pricing.

Expected Outcome

Create a separate "Save" button for offers which have already been consigned to an ACA. This will not trigger an automated email to the consignor offer email (as with the organic pipeline approvals), but will officially record the offer in Convection for data tracking purposes. Or re-purpose the existing "Accept Offer" button in the offer draft stage so that the offer can be officially recorded without being "sent" to the user.

Actual outcome
To circumvent this, our team has been using their own user accounts to send through the offer, and then going back later to update the submission with the consignor's email for tracking. Unfortunately this is error prone and unsustainable considering our team's bandwidth and cadence. This results in EOQ backlog editing, which means that there is insufficient and unreliable data throughout the quarter, which is problematic for us all.

### DIFF
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -28,6 +28,7 @@ class Offer < ApplicationRecord
   STATES = [
     DRAFT = 'draft',
     SENT = 'sent',
+    SAVED = 'saved',
     ACCEPTED = 'accepted',
     REJECTED = 'rejected',
     LAPSED = 'lapsed',
@@ -78,7 +79,7 @@ class Offer < ApplicationRecord
   end
 
   def reviewed?
-    !draft? && !sent? && !review?
+    !draft? && !sent? && !review? && !saved?
   end
 
   def locked?
--- a/app/views/admin/offers/_state_actions.html.erb
+++ b/app/views/admin/offers/_state_actions.html.erb
@@ -3,6 +3,9 @@
     <div class='overview-section offer-draft-actions'>
       <div class='single-padding-top'>
         <div>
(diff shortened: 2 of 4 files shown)


bhoggard:
@Gon4ar5 have you verified that a change to this state won't send an email?

Gon4ar5:
> @Gon4ar5 have you verified that a change to this state won't send an email?

Offer message send from here https://github.com/artsy/convection/blob/master/app/services/offer_service.rb#L51

But i can add some spec tommorow for clarity that it works https://github.com/artsy/convection/blob/master/app/services/offer_service.rb#L51

bhoggard:
Yes, I would feel more comfortable with a spec, to make sure we don't break it in the future. Thank you.

bhoggard (approved):

sarahshels:
Hi Barry and Ole, thank you so much for your work on this! So that I can best explain this utility to the team, can you please give me a bullet point of what action gets triggered by each button above?

PR #1109, opened 2021-07-23
fix: Remove extra offers spec file

olerichter00:
## Description

Removes unnecessary offers spec file that broke the deploy.

See [Slack Thread](https://artsy.slack.com/archives/C01B2P6LJUU/p1627002091303300) for context.


Gon4ar5:
Thanks, Ole

PR #1111, opened 2021-07-23
feat(CX-1712) Add new Withdrawn states to consignment

Gon4ar5:
Resolve [CX-1712](https://artsyproduct.atlassian.net/browse/CX-1712)

Screenshot 2021-07-23 at 14 45 37

# Description:

### What is the problem?
There is currently not a reliable way to reflect an unrealized consignment in convection. (unrealized being agreed but never signed the contract, became unresponsive, had to withdraw due to authenticity concerns, etc)

### Expected Outcome
Create a state in Convection to track unrealized GMV

### Actual outcome
The current process is to delete the estimate and commission field information so that the data doesn't skew overall consignments reporting. This is creating an incomplete data set for leads and the exec team to review as they can't see what didn't make it to auction. An easy fix would be to add "Withdrawn - Pre-launch" and "Withdrawn - Post-Launch" states to Convection so that these can be filtered easily in looker, and we don't lose the important data.

### DIFF
--- a/app/models/partner_submission.rb
+++ b/app/models/partner_submission.rb
@@ -23,7 +23,7 @@ class PartnerSubmission < ApplicationRecord
   has_many :offers, dependent: :destroy
   belongs_to :accepted_offer, class_name: 'Offer'
 
-  STATES = ['open', 'sold', 'bought in', 'canceled', 'withdrawn - Pre-Launch', 'withdrawn - Post-Launch'].freeze
+  STATES = ['open', 'sold', 'bought in', 'canceled', 'withdrawn - pre-launch', 'withdrawn - post-launch'].freeze
 
   scope :group_by_day, -> { group("date_trunc('day', notified_at) ") }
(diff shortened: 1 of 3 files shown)


app/models/partner_submission.rb
```diff
@@ -23,7 +23,7 @@ class PartnerSubmission < ApplicationRecord
   has_many :offers, dependent: :destroy
   belongs_to :accepted_offer, class_name: 'Offer'
 
-  STATES = ['open', 'sold', 'bought in', 'canceled'].freeze
+  STATES = ['open', 'sold', 'bought in', 'canceled', 'withdrawn - Pre-Launch', 'withdrawn - Post-Launch'].freeze
```
olerichter00, line 26:
```suggestion
  STATES = ['open', 'sold', 'bought in', 'canceled', 'withdrawn - pre-launch', 'withdrawn - post-launch'].freeze
```

olerichter00 (approved):
LGTM 👍

bhoggard (approved):

PR #1112, opened 2021-07-23
feat: remove global_constant file and rake task for it

Gon4ar5:
Since we have already deployed convection and executed a rake task for it, it makes sense to delete files that are no longer used. What do you think about this?


bhoggard (approved):

olerichter00:
Thanks @Gon4ar5 !

PR #1122, opened 2021-08-23
WIP - what would it be like to accept anonymous submission?

jonallured:
I took a stab today at what it might be like to accept anonymous consignment submissions. My approach was to create a singleton anonymous user and then substitute them in for a regular convection user during the create submission flow. To do this via the UI I added a checkbox to the submission form that hijacks things at the create action and sends in that anon user's gravity user id.

From there the flow is identical.

Note that this is far from complete and doesn't address how we'd allow a client like force to create these anonymous submissions. Still, I think this illustrates that overall the work is not too intense.

### DIFF
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -61,10 +61,15 @@ module Admin
     end
 
     def create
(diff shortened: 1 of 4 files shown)


jonallured:
Closing in favor of #1131

PR #1126, opened 2021-08-27
chore: Fix severe dependency updates

jacobherrington:
I'm manually updating some of the dependencies that have been flagged as security issues by GitHub.


Gemfile
```diff
@@ -4,14 +4,15 @@ source 'https://rubygems.org'
 
 ruby File.read('.ruby-version')
 
-gem 'rails', '6.1.3.2'
+gem 'rails', '6.1.4.1'
 
 gem 'pg'
-gem 'puma'
+gem 'puma', '~> 5.4.0'
```
jonallured, line 10: I'm curious about this pin - why not leave it off?

Gemfile
```diff
@@ -4,14 +4,15 @@ source 'https://rubygems.org'
...
 
 watt_gem_spec = { git: 'https://github.com/artsy/watt.git', branch: 'master' }
 gem 'watt', watt_gem_spec # artsy bootstrap
 
+gem 'addressable', '>= 2.8.0' # https://github.com/advisories/GHSA-jxhc-q857-3j6g
```
jonallured, line 15: I get that we need 2.8.0 in order to have a version without that vulnerability, but don't we get that even without this line? Just not sure why we need to include this gem here rather than relying on `bundle update` to get us on the latest version.

jonallured (commented):
Hey hey, thanks for the update PR! I'm curious about some of the Gemfile changes but this is wonderful, thank you!!

jonallured:
Oh and to answer your QA question, I would totally rely on the tests here. For more major upgrades I have occasionally cruised around staging or production as things roll out but I have a high confidence that if CI is green then it's good to deploy.

Gemfile
jacobherrington: No specific reason, I tend to like these more explicit call-outs in the gemfile just for the sake of people reading the file, but it's not necessary.

Gemfile
jacobherrington: It's not necessary, I just have a habit of being pretty specific about which versions of dependencies are in use. I went ahead and removed this line since the rest of the dependencies aren't quite so specific.

jacobherrington:
@jonallured Thanks for the review!

I updated the PR to be less explicit about versions. 👍

jonallured (approved):
LGTM!! 🎉

PR #1131, opened 2021-09-01
feat: Allow anonymous consignment submissions GRO-504

jonallured:
This PR supports the goal of allowing more anonymous consignment submissions. It's a spiritual successor to https://github.com/artsy/convection/pull/1122 but closer to reality than that spike.

## The Anonymous User

Taken from that other PR is the idea of a singleton user that we can attach these submission to. My approach to this was defining a class method on user called `anonymous` that uses Rails' `find_or_create_by` method to return or make the singleton user.

## A Special Param

Then in order to work with this user I'm now supporting a special param called `gravity_user_id`. When that is passed and is set to `anonymous`, then slightly different authentication logic and current user calculation are used. For the auth logic we look to the JWT being used and see if it's trusted. If it is, then cool, if not, then we fallback to the existing auth logic.

If the app is deemed trusted then we set the `current_user` to that anon user and march on.

## Updated Tests

To drive this out I started by looking at the create and update specs for the REST api endpoints. I did some organizing of what I found and then added my behavior to demonstrate how this can be used. It'll take a little more messing around to figure out how to generate and send the JWT token correctly from force but we're smart, we can figure it out!

https://artsyproduct.atlassian.net/browse/GRO-504

/cc @artsy/grow-devs @joeyAghion

### DIFF
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -2,7 +2,8 @@
 
 module Api
   class SubmissionsController < RestController
-    before_action :require_authentication
+    before_action :require_authentication, except: :create
+    before_action :require_app_or_auth, only: :create
     before_action :set_submission, only: %i[show update]
     before_action :require_authorized_submission, only: %i[show update]
 
@@ -37,6 +38,15 @@ module Api
     end
 
     private
+
+    def require_app_or_auth
+      if params[:gravity_user_id]
+        require_trusted_app
+        @current_user = User.anonymous.gravity_user_id
+      else
+        require_authentication
+      end
+    end
 
     def submission_params
       params.permit(
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -2,8 +2,8 @@
 
 module Api
   class SubmissionsController < RestController
-    before_action :require_authentication, except: :create
-    before_action :require_app_or_auth, only: :create
+    before_action :require_authentication, except: %i[create update]
+    before_action :require_app_or_auth, only: %i[create update]
     before_action :set_submission, only: %i[show update]
     before_action :require_authorized_submission, only: %i[show update]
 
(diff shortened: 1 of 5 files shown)


damassi (approved):
Looks good! And thanks for the additional cleanup.

joeyAghion:
Note that Gravity already offers the concept of anonymous, user-like data in the `AnonymousSession` class. The only requirement is that the client performing the anonymous interactions provides a globally unique identifier for it in `session_id`. This model is polymorphically associated with inquiries, and previously with e-commerce orders. I'd rather we use that or at least model anonymity separately from `User`s... it feels dangerous that every authentication- or authorization-related function may have to guard against the possibility of a special no-one user.

jonallured:
Interesting! I had not bumped into `AnonymousSession` before so I did some reading and poking around - I could see us iterating on this to use that modeling. I'm not super clear on the danger you're feeling @joeyAghion but initially the barrier is going to be api support. Convection is an odd-ball in that it does not use Kinetic and instead has it's own v2 and gravql clients. Unless I goofed my searches through Gravity I don't believe `AnonymousSession` is supported outside v1.

But that's something we could tackle - I'm just not sure it's worth the effort or if the risk warrants it. 🤷

joeyAghion:
The danger I'm imagining is around cases where a legitimate user _is_ still required, either as the authenticated visitor or an association. Locations that used to check `if current_user` would need to check `if current_user && current_user.gravity_user_id != 'anonymous'` or locations that require a [real] user id would need to add a similar guard.

jonallured:
Sorry I'm not getting it @joeyAghion! 😢 Could you give an example where there's a danger that would be mitigated by using Gravity's `AnonymousSession` model rather than Convection's `User#anonymous` singleton user?

joeyAghion:
Sorry for conflating 2 things:
1. modeling unauthenticated submissions in a first-class way, and
2. whether to use Gravity's existing `AnonymousSession` concept

I'm most concerned about (1), because a special ID is all that distinguishes those submissions.

iskounen:
> Could you give an example where there's a danger that would be mitigated by using Gravity's `AnonymousSession` model rather than Convection's `User#anonymous` singleton user?

You could use the consignment submission flow to upload photos to other anonymous submissions by guessing their IDs.

PR #1134, opened 2021-09-15
Fix node/husky set-up and add new pre-commit hook to scan for potential secrets (PLATFORM-3415)

joeyAghion:
This started out just aiming to add a secret-scanning hook, but Convection has dependencies on node and husky (and lint-staged) that weren't actually working so we (@artsy/platform-engineers) had to take a step back and upgrade and re-configure those. The first commit covers the revised set-up, after which the relatively simple next commit adds secret-scanning following the `brew` approach being evaluated by Gravity and Pulse.

https://artsyproduct.atlassian.net/browse/PLATFORM-3415


.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install lint-staged
```
jacobherrington, line 4: Did we decide if this was the proper way to run this?

We're not using the --no-install flag [in pulse](https://github.com/artsy/pulse/blob/adae00989a605c5b0b99006b9f16107270b51605/.husky/pre-commit#L4).

I think it makes sense to include this flag, but we might also want to add it to pulse.

jacobherrington (approved):
This is working locally for me!

I think commits will fail if people don't run `yarn install` after this commit. It might be a good idea to communicate that, but everything looks okay to me. 👍

.husky/pre-commit
joeyAghion: I was just following [their examples](https://typicode.github.io/husky/#/?id=locally-installed-binaries) which all seem to do this. Reading the related discussions (like https://github.com/zkat/npx/issues/198), it seems like the recommended way to do this for already-installed dependencies, as we wouldn't want this hook to modify `package.json` or `yarn.lock`.

In our mob, we tried removing it while debugging, but the problems ended up being unrelated. I.e., I think we probably _should_ add a similar flag to Pulse and elsewhere.

PR #1135, opened 2021-09-15
security: update artsy-auth and mitigate CVE-2015-9284

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3552

Following https://github.com/artsy/artsy-auth/pull/9, this PR updates artsy-auth to v0.2.0 to mitigate CVE-2015-9284.

`bundle update artsy-auth` was run, which updated `artsy-auth` and all its dependencies.


Gemfile.lock
```diff
@@ -502,7 +523,6 @@ DEPENDENCIES
   kaminari
   money
   neat (= 1.7.2)
-  omniauth (< 2)
```
artsyjian, line 505: This pin was placed in https://github.com/artsy/convection/pull/945. Removing, because `artsy-auth` v0.2.0 addresses the problem experienced with `omniauth` v2.

jonallured (approved):
LGTM!! 🎉

PR #1139, opened 2021-09-27
Fix(CX-1713): Fetch sale artworks method

Gon4ar5:
Just small fix fetch method
[CX-1713](https://artsyproduct.atlassian.net/browse/CX-1713)


olerichter00 (approved):
Looking good ⭐

PR #1154, opened 2021-11-08
feat[SWA-46]: Add new fields to user model

Gon4ar5:
Resolve [SWA-106](https://artsyproduct.atlassian.net/browse/SWA-106), also resolves [SWA-45](https://artsyproduct.atlassian.net/browse/SWA-45) and [SWA-46](https://artsyproduct.atlassian.net/browse/SWA-46)

Add new fields to User model
added the use of email, phone, name fields primarily from convection

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,8 +3,6 @@
 class User < ApplicationRecord
   include PgSearch::Model
 
-  validates :gravity_user_id, presence: true, uniqueness: true
-
   has_many :submissions, dependent: :nullify
   has_many :notes, dependent: :nullify
 
@@ -13,7 +11,7 @@ class User < ApplicationRecord
   def gravity_user
     if defined?(@gravity_user)
       @gravity_user
-    else
+    elsif gravity_user_id
       @gravity_user =
         gravity_user_id &&
           (
@@ -26,8 +24,16 @@ class User < ApplicationRecord
     end
   end
 
-  def name
-    gravity_user.try(:name)
+  def user_name
+    name || gravity_user.try(:name)
+  end
+
+  def user_email
+    email || user_detail&.email
+  end
+
+  def user_phone
+    phone || user_detail&.phone
   end
 
   def user_detail
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,16 +24,16 @@ class User < ApplicationRecord
     end
   end
 
-  def user_name
-    name || gravity_user.try(:name)
+  def name
+    self[:name] || gravity_user.try(:name)
   end
 
-  def user_email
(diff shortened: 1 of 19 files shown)


app/models/user.rb
```diff
@@ -26,8 +24,16 @@ def gravity_user
     end
   end
 
-  def name
-    gravity_user.try(:name)
+  def user_name
```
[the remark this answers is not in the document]
Gon4ar5: then ruby will not know the method or field of the object is called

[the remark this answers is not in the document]
iskounen: I agree that the `user_` prefix is redundant. I understand that there will be a clash between `name` the variable and `name` the method, but I have seen code like this to resolve that:

```ruby
def name
  @name || gravity_user.try(:name)
end
```

Does that not work in this class?

app/views/admin/submissions/_collector_info.html.erb
```diff
@@ -7,16 +7,16 @@
       </div>
     <% end %>
     <div>
-      <%= submission.user&.name %>
+      <%= submission.user.user_name %>
```
iskounen, line 10: You removed the `&` operator from `user`. Is it safe to remove that?

kiryl-zubarau (approved):
LGTM

app/models/user.rb
```diff
@@ -3,8 +3,6 @@
 class User < ApplicationRecord
   include PgSearch::Model
 
-  validates :gravity_user_id, presence: true, uniqueness: true
```
iskounen, line 6: @Gon4ar5 removing this check allows admins to manually create submissions without a user AND without a name/email/phone. That causes the `/submissions` and `/submission/[:id]` routes to fail with the following error:

```
NoMethodError: undefined method `user_detail' for #<Hyperclient::Resource:0x00007fc2055ed6d8>
  from hyperclient (1.0.1) lib/hyperclient/resource.rb:118:in `method_missing'
  from app/models/user.rb:40:in `user_detail'
  from app/models/user.rb:32:in `email'
  from app/views/admin/submissions/_submission.html.erb:22
  ...
  from app/views/admin/submissions/_submission.html.erb:3
 ...
  from app/views/admin/submissions/index.html.erb:91
```

We should patch this by reintroducing validation for `presence` if name/email/phone are not provided.

iskounen: @Gon4ar5 we removed the `uniqueness` check here, but we also need to remove that from the [database schema](https://github.com/artsy/convection/blob/ae17e282c887d75e4d86d5ab83e6d43a8e14b7df/db/schema.rb#L235-L237).

PR #1156, opened 2021-11-11
fix[SWA-122]: Displaying the correct cataloguer name

Gon4ar5:
Resolve [SWA-122](https://artsyproduct.atlassian.net/browse/SWA-122)

### Description:
Whenever any cataloguer is selected, it defaults to Emily's name, which isn't on the list

37f85920-62b4-478c-b51d-62b001776544

------------
### Why?
We save the gravity_user_id in the cataloguer field, but when displaying it on the view, we looked for cataloguer.name by id

Screen Shot 2021-11-11 at 3 28 04 PM


AlicanAkyuz (approved):
lgtm 🚀 

and thanks for picking this up immediately 🙏

PR #1163, opened 2021-11-16
feat[SWA-104]: Make the create-submission mutation public

Gon4ar5:
Resolve [SWA-104](https://artsyproduct.atlassian.net/browse/SWA-104)

- [x] Remove validation logic from CreateSubmissionResolver
- [x] Prevent SubmissionService from throwing an exception when a Gravity ID is not provided
- [x] Change the gravity_user_id field on the user model to be nullable
- [x] Change the user ID field to be optional in the submission GraphQL type
- [x] Change the user ID argument to be optional in the create-submission mutation
- [x] Add a session ID field to the submission model
- [x] Add a session ID field to the submission GraphQL type
- [x] Add a session ID argument to the create-submission mutation


kiryl-zubarau (approved):
LGTM

PR #1164, opened 2021-11-16
[feat]: SWA-126 - Allow admins to submit non-target-supply artists

AlicanAkyuz:
### Description 

This PR implements: 
-  do not automatically reject any submissions created within the Convection panel by our admins, even the artist is non-target supply or the user in the submission is not an admin
- keeps automatically rejecting non-target supply artist submissions made by a user themselves via web or mobile

Ticket: [SWA-126](https://artsyproduct.atlassian.net/browse/SWA-126) (see also the comments)

### Video 
- First 2 submissions are for a non-target supply artist within Convection, one by a user (alican) and the other is by an admin (Alican Akyuz)
- 3rd and 4th submissions are via a web, one by target supply artist and the other is not

https://user-images.githubusercontent.com/42584148/142196703-0a88baa1-e4f4-48ea-8f7b-20ea4bf11dbb.mov

### DIFF
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -66,7 +66,8 @@ module Admin
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          submission_params[:user_id]
+          submission_params[:user_id],
+          authorized_artsy_token?(session[:access_token])
         )
       redirect_to admin_submission_path(@submission)
     rescue SubmissionService::SubmissionError => e
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -66,8 +66,7 @@ module Admin
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          submission_params[:user_id],
-          authorized_artsy_token?(session[:access_token])
+          submission_params[:user_id]
         )
       redirect_to admin_submission_path(@submission)
     rescue SubmissionService::SubmissionError => e
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -15,15 +15,19 @@ class SubmissionService
       ]
       user = User.find_or_create_by!(gravity_user_id: gravity_user_id)
       create_params = submission_params.merge(user_id: user.id)
-      final_params =
+      create_params =
         create_params.merge(
-          reject_if_non_target_supply_artist(submission_params[:artist_id])
+          reject_non_target_supply_artist_by_non_admin(
+            submission_params[:artist_id],
+            user,
+          ),
         )
-      submission = Submission.create!(final_params)
 
-      if final_params[:state] == 'rejected'
+      submission = Submission.create!(create_params)
+
+      if create_params[:state] == 'rejected'
         delay_until(
-          Convection.config.rejection_email_minutes_after.minutes.from_now
+          Convection.config.rejection_email_minutes_after.minutes.from_now,
         ).deliver_rejection_notification(submission.id)
       end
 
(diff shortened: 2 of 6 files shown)


kiryl-zubarau (approved):
LGTM

AlicanAkyuz:
@kiryl-zubarau @Gon4ar5 I changed the code after talking to product, and updated the video. This PR is ready to go from my side.

kiryl-zubarau (approved):

app/services/submission_service.rb
```diff
@@ -8,22 +8,26 @@ class SubmissionError < StandardError
...
-        )
-      submission = Submission.create!(final_params)
 
-      if final_params[:state] == 'rejected'
+      unless is_convection
+        create_params =
+          create_params.merge(
+            reject_non_target_supply_artist(submission_params[:artist_id]),
+          )
+      end
```
Gon4ar5, line 24:
```suggestion
      unless is_convection
        create_params.merge!(reject_non_target_supply_artist(submission_params[:artist_id]))
      end
```

app/controllers/admin/submissions_controller.rb
```diff
@@ -66,7 +66,8 @@ def create
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          submission_params[:user_id]
+          submission_params[:user_id],
+          authorized_artsy_token?(session[:access_token])
```
Gon4ar5, line 70: The ticket states that admins should be able to create submissions for non-target artists, does this only apply to admins? If so, is it easier to check against `AdminUser.exists? (gravity_user_id: user)`

app/services/submission_service.rb
```diff
@@ -8,22 +8,26 @@ class SubmissionError < StandardError
   end
 
   class << self
-    def create_submission(submission_params, gravity_user_id)
+    def create_submission(submission_params, gravity_user_id, is_convection)
```
Gon4ar5, line 11: You can set default value, and don't pass false elsewhere
```suggestion
    def create_submission(submission_params, gravity_user_id, is_convection = false)
```

app/controllers/admin/submissions_controller.rb
AlicanAkyuz: I think we are not interested in whether user is an admin or not, because sometimes a collector (user) asks an Artsy admin to create a submission for them. In such a case, our admin would put collector as user in the submission, and user wouldn't be an admin, but we still want to allow submission with non-target supply artist. 

Basically, what we want: 
- do not automatically reject, if submission created in Convection (even if submission's user is not an admin)
- automatically reject, if submission is created by user via web or mobile and it's a non-target supply artist

AlicanAkyuz: `authorized_artsy_token?(session[:access_token])`

⬆️  this works in pratice, but what do you think? is there a better way of doing it?

app/services/submission_service.rb
AlicanAkyuz: when I set it as `is_convection = false`, the CI complains: `Use keyword arguments when defining method with boolean argument`. I guess it wants me to use `is_convection: false`, but that behaves differently.  

do you know how to overcome this?

AlicanAkyuz: here: https://app.circleci.com/pipelines/github/artsy/convection/3152/workflows/5201dd79-e92e-4ec5-a2a3-20d8de795de4/jobs/3787

app/controllers/admin/submissions_controller.rb
Gon4ar5: That is, we want to do this only for submissions who come from outside? Maybe then it would be better to add this logic to create_submission_resolver?

AlicanAkyuz: lemme try!

app/services/submission_service.rb
Gon4ar5: Oh, rubocop 
u have 3 options: 
1) disable OptionalBooleanParameter 
2) Use `is_convection: false`, but call the method with the third parameters explicitly specifying the third parameter, such as `SubmissionService.create_submission (params, 'userid', is_convection: true)` 
3) Change it to nil `create_submission(submission_params, gravity_user_id, is_convection = nil)`

PR #1173, opened 2021-11-20
fix: an email parameter is being assigned to a phone field when creating anonymous submissions

iskounen:
I noticed that the value of `user_email` was being assigned to the `phone` field of a user record. This PR fixes that by assigning the value of `user_phone` instead.


Gon4ar5 (approved):
Thanks for the fix, LGTM

PR #1175, opened 2021-11-20
feat: save contact information for authenticated submissions

iskounen:
This PR updates the `create_submission` service method to save contact information (name, email, phone) for authenticated submissions.

In the new web submission flow, users are asked to enter their contact information. We are currently saving this information for anonymous submissions, but we would also like to save it for authenticated submissions.

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -19,7 +19,14 @@ class SubmissionService
       ]
       user =
         if gravity_user_id
-          User.find_or_create_by!(gravity_user_id: gravity_user_id)
+          convection_user =
+            User.find_or_create_by!(gravity_user_id: gravity_user_id)
+          convection_user.update(
+            name: submission_params[:user_name],
+            email: submission_params[:user_email],
+            phone: submission_params[:user_phone]
+          )
+          convection_user
         else
           User.create!(
             name: submission_params[:user_name],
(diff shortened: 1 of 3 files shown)


app/services/submission_service.rb
```diff
@@ -19,7 +19,14 @@ def create_submission(
       ]
       user =
         if gravity_user_id
-          User.find_or_create_by!(gravity_user_id: gravity_user_id)
+          convection_user =
```
Gon4ar5, line 22: Great, but if-else does almost the same thing, can we refactor something like:
```
      user = User.find_or_create_by!(gravity_user_id: gravity_user_id) if gravity_user_id
      user ||= User.create!
      user.update(
        name: submission_params[:user_name],
        email: submission_params[:user_email],
        phone: submission_params[:user_phone],
        session_id: submission_params.delete(:session_id)
      )
      create_params = submission_params.merge(user_id: user.id)
```

Gon4ar5: Another question is whether it is necessary to check this data for presence? I mean, when creating a submission, this data will overwrite the already existing data about the user, and even if it does not come, they will still overwrite it.

iskounen: Thanks for noticing that there is a refactoring opportunity here! I agree that (1) the if-else is not necessary, and (2) we won't need to find existing users once there is a 1:1 relationship between submission and user records.

I will close this PR so that you can do this refactoring in #1179

PR #1176, opened 2021-11-20
fix: require a gravity user ID for submissions when contact information is not provided

iskounen:
This PR addresses an issue where admins were able to manually create submissions without a user in the admin UI. This caused the submission list page (`/submissions`) to crash when it tried to render submissions with neither a `gravity_user_id` nor any contact information (name, email, or phone).

This was resolved by re-introducing validation for the `presence` of a `gravity_user_id`, on the condition that it is only required when contact information is not provided.

Screen Shot 2021-11-22 at 9 40 43 AM

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,25 +3,26 @@
 class User < ApplicationRecord
   include PgSearch::Model
 
+  validates :gravity_user_id, presence: true, unless: :contact_information?
+
   has_many :submissions, dependent: :nullify
   has_many :notes, dependent: :nullify
 
   pg_search_scope :search, against: :email, using: { tsearch: { prefix: true } }
 
   def gravity_user
-    if defined?(@gravity_user)
-      @gravity_user
-    elsif gravity_user_id
-      @gravity_user =
-        gravity_user_id &&
-          (
-            begin
-              Gravity.client.user(id: gravity_user_id)._get
-            rescue Faraday::ResourceNotFound
-              nil
-            end
-          )
-    end
+    return @gravity_user if defined?(@gravity_user)
+    return nil unless gravity_user_id
+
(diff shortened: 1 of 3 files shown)


iskounen:
This PR is a WIP because I am still able to create a manual submission without a user in the admin screens. The submissions created in this way are no longer crashing the submission list page (`/submissions`) but I expected the form to fail validation. Instead, a user with the following contact information was assigned to the submission: user, user@example.com, 555-5555.

app/models/user.rb
```diff
@@ -3,25 +3,26 @@
...
+
+    @gravity_user =
+      gravity_user_id &&
+        (
+          begin
+            Gravity.client.user(id: gravity_user_id)._get
+          rescue Faraday::ResourceNotFound
+            nil
+          end
+        )
```
iskounen, line 25: comment: This is a little refactor that I did to clarify the values that this method can return.

AlicanAkyuz (approved):
LGTM! 

I've pulled and run the branch, and I see the user can't be blank error when a submission without a user is attempted.

app/models/user.rb
```diff
@@ -49,4 +50,8 @@ def unique_code_for_digest
   def self.anonymous
     User.find_or_create_by(gravity_user_id: 'anonymous')
   end
+
+  def contact_information?
+    !(self[:name].blank? || self[:email].blank? || self[:phone].blank?)
```
Gon4ar5, line 55: Just a suggestion: could replace this with something like this:
```suggestion
    self[:name].present? && self[:email].present? && self[:phone].present?
```
This negation combined with unless on top is a bit confusing😆

Gon4ar5 (approved):
LGTM!
Also work for me

PR #1179, opened 2021-11-22
feat[SWA-131]: Create a one-to-one relationship submissions/users

Gon4ar5:
Related [SWA-131](https://artsyproduct.atlassian.net/browse/SWA-131)

1) Changed relation from has_many to has_one
1.1) Also changed user.submissions to user.submission
2) Added rake task which creates new user, and reassign submission on him if some submission has the same user_id
3) Changed user creation in submission_service, now we don't try to find an existing one, but immediately create a new one

### DIFF
--- a/app/controllers/admin/consignments_controller.rb
+++ b/app/controllers/admin/consignments_controller.rb
@@ -22,9 +22,9 @@ module Admin
 
       if filtering_by_assigned_to?
         matching_consignments =
-          matching_consignments.joins(:submission).where(
-            'submissions.assigned_to' => params[:assigned_to]
-          )
+          matching_consignments
+            .joins(:submission)
+            .where('submissions.assigned_to' => params[:assigned_to])
       end
 
       if term.present?
@@ -39,13 +39,14 @@ module Admin
       if params[:user].present?
         user = User.find(params[:user])
         matching_consignments =
-          matching_consignments.where(submission: user.submissions)
+          matching_consignments.where(submission: user.submission)
       end
 
       if params[:artist].present?
-        matching_consignments = matching_consignments.
-          joins(:submission).
-          where(submissions: { artist_id: params[:artist]})
+        matching_consignments =
+          matching_consignments
+            .joins(:submission)
+            .where(submissions: { artist_id: params[:artist] })
       end
 
       if params[:state].present?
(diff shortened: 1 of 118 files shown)


app/controllers/admin/consignments_controller.rb
```diff
@@ -39,13 +39,14 @@ class ConsignmentsController < ApplicationController
       if params[:user].present?
         user = User.find(params[:user])
         matching_consignments =
-          matching_consignments.where(submission: user.submissions)
+          matching_consignments.where(submission: user.submission)
```
Gon4ar5, line 42: changed in this file only here

app/controllers/admin/offers_controller.rb
```diff
@@ -40,22 +46,22 @@ class OffersController < ApplicationController
 
       if params[:user].present?
         user = User.find(params[:user])
-        matching_offers = matching_offers.where(submission: user.submissions)
+        matching_offers = matching_offers.where(submission: user.submission)
```
Gon4ar5, line 49: changed in this file only here

db/migrate/20211121122634_change_gravity_user_id_to_not_unique.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
Gon4ar5, line 1: already merge PR with this file

app/models/user.rb
```diff
@@ -42,8 +42,19 @@ def user_detail
     nil
   end
 
+  def user_submissions
```
kiryl-zubarau, line 45: Can we use just `submissions`?

app/models/user.rb
```diff
@@ -42,8 +42,19 @@ def user_detail
...
+    users =
+      if gravity_user_id
+        User.where(gravity_user_id: gravity_user_id)
+      else
+        User.where(name: name, email: email, phone: phone)
```
kiryl-zubarau, line 50: @iskounen Do we want to try to identify users by using such fields combination or just ignore such submissions?

Gon4ar5: @iskounen @kiryl-zubarau I don't think this is very important, since I use this method only to display the number of submissions of a given user on the page in the Convection admin
Screen Shot 2021-11-23 at 6 57 11 PM

iskounen: I think it's OK to display the number of submissions made by a user like this. This number will not always be accurate for anonymous submissions, but I think that it is still valuable for admins to see an approximation of how many submissions have been made by someone with the same name + email + phone.

lib/tasks/submissions.rake
```diff
@@ -5,3 +5,32 @@ task :exchange_assigned_to_real_user => :environment do
...
   end
 end
+    
+task correction_seq_id: :environment do
+  ActiveRecord::Base.connection.tables.each do |t|
+    ActiveRecord::Base.connection.reset_pk_sequence!(t)
+  end
+end
+
+task :rebase_user_submission => :environment do
```
iskounen, line 15: If possible, I'd like to run this task together on staging once this PR is merged and ready.

iskounen (approved):

kiryl-zubarau (approved):

AlicanAkyuz (approved):
lgtm 🥇

iskounen:
**blocker:** Given some feedback I received offline about this change, I've reconsidered the decision to store contact information on the user record. We will pause on this PR while @Gon4ar5 prepares a separate PR that:

1. Stores contact information provided with each submission on the `Submission` table.
2. Maintains the one-to-many relationship between a `User` and `Submission` records.
3. Makes the `user` field on a `Submission` optional.
4. Admin screens and Mailers default to the contact information on the `Submission` record.

PR #1180, opened 2021-11-23
[Feat]: SWA-100 - Update addAsset mutations

AlicanAkyuz:
### Description

Ticket: [SWA-100](https://artsyproduct.atlassian.net/browse/SWA-100)

This PR adds additional validation to addAsset mutations. The validation is done though sessionId associated with a user so that a user cannot add assets to another user's submission. 

### Implementation 

- matching_user method is moved from update_submission_resolver to the base resolver 
- matching_user is then utilised in add_asset and add_assets to_submission_resolver
- sessionId is added to those mutations as an argument

### DIFF
--- a/spec/requests/submission_spec.rb
+++ b/spec/requests/submission_spec.rb
@@ -14,7 +14,7 @@ describe 'Submission Flow' do
     expect(NotificationService).to receive(:post_submission_event).once
   end
 
-  it 'Completes 3' do
+  it 'Completes a submission from Tokyo' do
     stub_gravity_root
     stub_gravity_user
     stub_gravity_artist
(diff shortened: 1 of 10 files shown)


spec/requests/submission_spec.rb
```diff
@@ -14,7 +14,7 @@
     expect(NotificationService).to receive(:post_submission_event).once
   end
 
-  it 'Completes 3' do
+  it 'Completes a submission from Tokyo' do
```
AlicanAkyuz, line 17: this test description was changed recently by mistake, I just changed it to correct text. not directly related to the task.

kiryl-zubarau (approved):
🚀

PR #1181, opened 2021-11-24
Prettify the Repo

AlicanAkyuz:
### Description 

Ticket: [SWA-123](https://artsyproduct.atlassian.net/browse/SWA-123) 

The aim here is to lint the repo and sync local/pre-commit linters so that we don't get unrelated diffs in our PRs. An example of such diffs can be seen in [this PR](https://github.com/artsy/convection/pull/1179) where engineers put comments to highlight the actual changes. 

### Implementation

I have run `yarn run prettier --write` on the repo, but then the pre-commit hook fails at the step of `bundle exec rubocop -a`. There seems to be a conflict between prettier and rubocop configs. I had to manually remove `lint-staged` command from  husky's `pre-commit` file in order to commit the changes, but this is certainly not ideal. 

What do people think moving forward here?


Gon4ar5 (commented):
Maybe we should just turn off prettify😆, fix rubocop and merge this?
I don’t know how good such a solution is, but having two code analyzers that have conflicts with each other is also not a good idea.😅

app/controllers/admin/admin_users_controller.rb
```diff
@@ -67,9 +67,9 @@ def admin_user_params
     end
 
     def authorize_user!
-      unless super_admin_user? @current_user
-        raise ApplicationController::NotAuthorized
-      end
+      return if super_admin_user? @current_user
+
+      raise ApplicationController::NotAuthorized
```
kiryl-zubarau: The place where I made a manual fix. Other problems were auto resolved by `bundle exec rubocop -a`

AlicanAkyuz:
@Gon4ar5 @kiryl-zubarau this is ready for final review.

Gon4ar5 (approved):
LGTM

kiryl-zubarau (approved):

PR #1184, opened 2021-11-30
[feat]: Swa-134 - created_by column on submissions

AlicanAkyuz:
### Description 

Ticket: [SWA-134](https://artsyproduct.atlassian.net/browse/SWA-134) 

Convection admins might create submissions on behalf of users, as opposed to users creating their own submissions via Force or Eigen. When evaluating a submission, it is useful to know the creator of the submission. Therefore this PR: 
- adds a new column to Submission in db, called `created_by` 
- sets `created_by` to nil for all user submissions
- sets `created_by` to the email address of the admin, when submission is created by an admin
- displays Created By under Collector Info in Submission view 
- displays Created By in Submissions list view

#### List View  
Screenshot 2021-11-30 at 15 25 20

#### Submission View (created_by is nil)
Screenshot 2021-11-30 at 15 25 36

#### Submission View (created_by is admin email)
Screenshot 2021-11-30 at 15 24 03

### DIFF
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -66,7 +66,8 @@ module Admin
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          submission_params[:user_id]
+          submission_params[:user_id],
+          is_admin: AdminUser.exists?
         )
       redirect_to admin_submission_path(@submission)
     rescue SubmissionService::SubmissionError => e
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -67,7 +67,8 @@ module Admin
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
           submission_params[:user_id],
-          is_admin: AdminUser.exists?
+          is_admin: AdminUser.exists?,
+          current_user: @current_user
         )
       redirect_to admin_submission_path(@submission)
     rescue SubmissionService::SubmissionError => e
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -67,7 +67,6 @@ module Admin
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
           submission_params[:user_id],
-          is_admin: AdminUser.exists?,
           current_user: @current_user
         )
       redirect_to admin_submission_path(@submission)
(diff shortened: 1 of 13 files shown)


app/controllers/admin/submissions_controller.rb
```diff
@@ -66,7 +66,9 @@ def create
       @submission =
         SubmissionService.create_submission(
           submission_params.merge(state: 'submitted'),
-          submission_params[:user_id]
+          submission_params[:user_id],
+          is_admin: AdminUser.exists?,
```
Gon4ar5, line 70: I don't think it needs to be passed as a parameter
you are already passing current_user and by it you can determine whether the user is an admin
`AdminUser.exists?(gravity_user_id: current_user)`

db/migrate/20211130141653_remove_created_by_from_submissions.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveCreatedByFromSubmissions < ActiveRecord::Migration[6.1]
```
Gon4ar5, line 1: Why do we need a remove migration?

AlicanAkyuz: I first created the new column in camelCase, as `createdBy`, then removed it and added `created_by`. 

but I guess we can remove this migration now?

app/services/submission_service.rb
```diff
@@ -30,6 +31,12 @@ def create_submission(
       user.session_id = submission_params.delete(:session_id)
       create_params = submission_params.merge(user_id: user.id)
 
+      if AdminUser.exists?(gravity_user_id: current_user)
+        create_params.merge!(
+          created_by: User.find_by(gravity_user_id: current_user).email
+        )
+      end
```
Gon4ar5, line 38:
```suggestion
      create_params = submission_params.merge(
        user_id: user.id,
        created_by: User.find_by(gravity_user_id: current_user)&.email
      )
```

Gon4ar5: oh, do we save the user's email only if he is one of the admins?

Gon4ar5: if yes - undo my changes :)

Gon4ar5 (approved):
LGTM

app/services/submission_service.rb
AlicanAkyuz: oh hey yes, we want to save the admin's email only to the created_by field, and leave it nil if the operator is not an admin

PR #1185, opened 2021-12-01
Feat[SWA-131]: Prioritize using contact info from submission

Gon4ar5:
Related [SWA-131](https://artsyproduct.atlassian.net/browse/SWA-131)

The main goals of this PR:
  - Remove name, phone and session_id fields from user
  - Prioritize using email, name and phone from submission instead of gravity requests

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1523,6 +1523,16 @@ enum SubmissionSort {
   CREATED_AT_DESC
 
   """
+  sort by created_by in ascending order
+  """
+  CREATED_BY_ASC
+
+  """
+  sort by created_by in descending order
+  """
+  CREATED_BY_DESC
+
+  """
   sort by currency in ascending order
   """
   CURRENCY_ASC
@@ -1811,6 +1821,16 @@ enum SubmissionSort {
   sort by reminders_sent_count in descending order
   """
   REMINDERS_SENT_COUNT_DESC
+
+  """
+  sort by session_id in ascending order
+  """
+  SESSION_ID_ASC
+
+  """
+  sort by session_id in descending order
+  """
+  SESSION_ID_DESC
 
   """
   sort by signature in ascending order
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1523,16 +1523,6 @@ enum SubmissionSort {
   CREATED_AT_DESC
 
   """
-  sort by created_by in ascending order
-  """
-  CREATED_BY_ASC
-
-  """
-  sort by created_by in descending order
-  """
-  CREATED_BY_DESC
-
-  """
   sort by currency in ascending order
   """
   CURRENCY_ASC
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -18,7 +18,7 @@ class UserMailer < ApplicationMailer
               submission_id: submission.id
             }
     mail(
-      to: user.email,
+      to: submission.email,
       subject: 'Thank you for submitting your artwork to Artsy',
       bcc: Convection.config.bcc_email_address
     )
@@ -37,7 +37,7 @@ class UserMailer < ApplicationMailer
             unique_args: {
               submission_id: submission.id
             }
-    mail to: user.email, subject: "You're Almost Done"
+    mail to: submission.email, subject: "You're Almost Done"
   end
 
   def second_upload_reminder(submission:, user:)
@@ -53,7 +53,7 @@ class UserMailer < ApplicationMailer
             unique_args: {
               submission_id: submission.id
             }
-    mail to: user.email,
+    mail to: submission.email,
          subject: 'Artsy Consignments - complete your submission'
   end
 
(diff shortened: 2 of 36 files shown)


_schema.graphql
```diff
@@ -1522,6 +1522,16 @@ enum SubmissionSort {
...
+  """
+  sort by created_by in ascending order
+  """
+  CREATED_BY_ASC
+
+  """
+  sort by created_by in descending order
+  """
+  CREATED_BY_DESC
+
```
AlicanAkyuz, line 1534: do you know why created_by sorting is added in this PR? asking because this is implemented in another PR which is not merged to master yet.

app/models/user.rb
```diff
@@ -26,15 +26,15 @@ def gravity_user
   end
 
   def name
-    self[:name] || gravity_user.try(:name)
+    gravity_user.try(:name)
   end
```
AlicanAkyuz, line 30: do we want to get user name from gravity, even if user provides a (potentially different) name in contact information?

db/schema.rb
```diff
@@ -217,6 +218,8 @@
     t.string 'cataloguer'
     t.string 'user_name'
     t.string 'user_phone'
+    t.string 'created_by'
```
AlicanAkyuz, line 221: here also, `created_by` is added and I'm thinking why this might be the case 🤔 maybe I'm missing something here.

app/mailers/user_mailer.rb
```diff
@@ -37,7 +37,7 @@ def first_upload_reminder(submission:, user:)
             unique_args: {
               submission_id: submission.id
             }
-    mail to: user.email, subject: "You're Almost Done"
+    mail to: submission.email, subject: "You're Almost Done"
   end
 
   def second_upload_reminder(submission:, user:)
```
AlicanAkyuz, line 43: since none of the functions use `user.email `anymore in this file, do you think it'd make sense to remove `user:` param from those functions?

app/models/user.rb
iskounen: https://github.com/artsy/convection/blob/dd1a2571530faaca4a6d0ed50cb9bb52bcfd7b0a/app/models/submission.rb#L192-L194

@AlicanAkyuz, no we do not want to get user details from Gravity if a user provided contact information in the 3rd step of the new submission flow. The change above is what will make that happen for submissions.

app/mailers/user_mailer.rb
```diff
@@ -156,7 +159,7 @@ def other_submission_rejected(submission:, user:, artist:)
             unique_args: {
               submission_id: submission.id
             }
-    mail(to: user.email, subject: 'An update about your submission')
+    mail(to: submission.email, subject: 'An update about your submission')
   end
 
   def offer(offer:, artist:, user:)
```
iskounen, line 165: **suggestion:** Can we remove the `user` parameter from this method? `OfferService` passes the user associated with an offer's submission to this method, which we can get from the `offer` parameter.

https://github.com/artsy/convection/blob/dd1a2571530faaca4a6d0ed50cb9bb52bcfd7b0a/app/services/offer_service.rb#L152

_schema.graphql
Gon4ar5: This happened due to the fact that I was pulling your branch and running the migration and not rolling it back

Gon4ar5: that is, in my local db this field exists, therefore it is displayed in the schema

app/services/submission_service.rb
```diff
@@ -20,14 +20,11 @@ def create_submission(
       user =
         if gravity_user_id
           User.find_or_create_by!(gravity_user_id: gravity_user_id)
+        elsif contact_information?(submission_params)
+          User.anonymous
```
iskounen, line 24: **issue:** I prefer that we set the user to `nil` instead of assigning the anonymous user record to a submission. Are there any issues with having a `nil` user?

app/models/user.rb
Gon4ar5: I'm trying to get info from the [submission](https://github.com/artsy/convection/blob/dd1a2571530faaca4a6d0ed50cb9bb52bcfd7b0a/app/models/submission.rb#L193), but if it's not there, then I call the name method from the user, which takes info from gravity

AlicanAkyuz: understood, thank you.

AlicanAkyuz (approved):

app/services/submission_service.rb
```diff
@@ -22,13 +22,8 @@ def create_submission(
         if gravity_user_id
           User.find_or_create_by!(gravity_user_id: gravity_user_id)
         else
-          User.create!(
-            name: submission_params[:user_name],
-            email: submission_params[:user_email],
-            phone: submission_params[:user_phone]
-          )
+          User.create!
```
iskounen, line 25: I was thinking that we wouldn't even create a user for anonymous users and keep the `user_id` field on the submission as `nil`.

iskounen (approved):

PR #1188, opened 2021-12-07
Fix[SWA-160]: Added the correct counter for the number of submissions

Gon4ar5:
Resolve [SWA-160](https://artsyproduct.atlassian.net/browse/SWA-160)

Fix submission counter, cause he didn't correctly count the number of anonymous user submissions

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -196,4 +196,10 @@ class Submission < ApplicationRecord
   def phone
     user_phone || user&.phone
   end
+
+  def count_submissions_of_user
+    return Submission.where(user_email: user_email).count if user.nil?
+
+    user.submissions.count
+  end
 end
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -198,7 +198,7 @@ class Submission < ApplicationRecord
   end
 
   def count_submissions_of_user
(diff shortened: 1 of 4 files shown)


app/models/submission.rb
```diff
@@ -196,4 +196,10 @@ def name
   def phone
     user_phone || user&.phone
   end
+
+  def count_submissions_of_user
+    return Submission.where(user_email: user_email).count if user.nil?
```
Gon4ar5, line 201: What do you think about finding submissions exclusively by the same emails?

kiryl-zubarau: I suppose It should also ignore non anonymous submissions.

Gon4ar5: If our submission is anonymous, then a user is not created for it

Gon4ar5: Resolved

kiryl-zubarau: I mean that if the submission is anonymous you should search among other anonymous submissions instead of all of their

Gon4ar5: I understand, but now it doesn't matter anymore :)

iskounen (approved):

PR #1190, opened 2021-12-09
feat[SWA-164] Add asset size and name on convection

Gon4ar5:
Resolve [SWA-164](https://artsyproduct.atlassian.net/browse/SWA-164)
Add new fields to Asset model and mutations


iskounen (approved):
suggestion: I think that `filename` would be a bit more descriptive than `name`

AlicanAkyuz (approved):

kiryl-zubarau (approved):

PR #1199, opened 2021-12-15
[feat]: Create RemoveAssetFromConsignmentSubmission mutation

AlicanAkyuz:
### Description 

Within the context of [SWA-166](https://artsyproduct.atlassian.net/browse/SWA-166), this PR adds a new mutation to in order for clients to be able to make requests to remove an asset from a submission.

#### Video 

https://user-images.githubusercontent.com/42584148/147102418-84df9a30-11f1-46d7-b90b-0cb9bd02590b.mov


app/graphql/mutations/remove_asset_from_consignment_submission.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Mutations
+    class RemoveAssetFromConsignmentSubmission < Mutations::BaseMutation
+        argument :gemini_token, String, required: true
+        argument :submissionID, ID, required: true
+        argument :sessionID, String, required: false
+        argument :asset_type, String, required: false
+        argument :filename, String, required: false
+        argument :size, String, required: false
```
kiryl-zubarau, line 10: Looks like there are a few unnecessary parameters: asset_type, filename, size. I also think we can use assetID instead of gemini_token and submissionID, in this case, we will load the exact asset then load their submission to validate the user, and if all is good we will be able to remove the asset.

kiryl-zubarau: @Gon4ar5 Could you please help with that?

kiryl-zubarau (approved):

PR #1200, opened 2021-12-16
fix: filtering submissions by users doesn't work for all email addresses

iskounen:
Paired with @AlicanAkyuz 

This PR fixes a bug where admins were not able to filter the list off submissions (`/admin/submissions`) by a collector's email address. This was only happening for collectors who submitted artworks after we made the change to store contact information on the submission model instead of the user model.

The solution was to include contact information as a search criteria for users, and remove any duplicate users that may be found with the regular user search.


AlicanAkyuz (approved):

Gon4ar5 (approved):
LGTM

PR #1203, opened 2021-12-17
feat[SWA-175]:Add ability to view all user submissions from submission count link

Gon4ar5:
Resolve [SWA-175](https://artsyproduct.atlassian.net/browse/SWA-175)

Made from the text 'No. submissions count' a link that redirect to the page with user submissions

#Update - Now it works like this: if a submission has a user, we display all submissions of this user, if not, then all submissions with the same user_email

https://user-images.githubusercontent.com/21379857/148948778-2dec5d1f-2d2c-4dea-bab5-14021d0cc50a.mov


AlicanAkyuz (approved):
looks good to me 🚀  

ps: if you wanted to highlight the link, I think adding `class: 'sidebar-link'` would suffice.

app/views/admin/submissions/_collector_info.html.erb
```diff
@@ -9,8 +9,8 @@
     <div>
       <%= submission.name %>
     </div>
-    <div>
-      <%= submission.count_submissions_of_user %> submissions made
+    <div class="sidebar-link-2" >
+      <%= link_to "#{submission.count_submissions_of_user} submissions made", admin_submissions_path(user: submission.user_id) %>
```
iskounen, line 13: **question (blocking):** What will happen if an admin clicks on this link for an anonymous submission?

iskounen (changes requested):

app/views/admin/submissions/_collector_info.html.erb
Gon4ar5: anonymous submission also had a user_id, that is, it will select all submissions with this user_id, if the user_id is still not specified, then it will be redirected to an empty page with submissions
Screen Shot 2021-12-17 at 4 02 44 PM

iskounen: Anonymous submissions shouldn't have a `user_id`, so I think that admins will always see a blank page if they click this link from an anonymous submission.

If 10 submissions were made with the same email address, then an admin would expect to see 10 submissions in the list when they click this link. However, if 1 of those 10 submissions is an anonymous submission, and they click this link from that anonymous submission, then they will see an empty list.

Gon4ar5: I double-checked and this is a really great note, I will think about how to fix it

spec/models/submission_spec.rb
```diff
@@ -254,7 +254,7 @@
       end
 
       it 'return the count of user submissions found by user_email eq 1' do
-        expect(submission.count_submissions_of_user).to eq 1
+        expect(submission.count_submissions_of_user).to eq 2
```
iskounen, line 257: **nitpick:**

1.  The label of this test case should say "... eq 2"
2. The label of the test case above this should say "... when 1 is deleted" to differentiate them.

app/assets/stylesheets/locals.css.scss
```diff
@@ -51,6 +51,12 @@ $green: #0eda83;
   text-decoration: underline;
 }
 
+.sidebar-link-2 {
```
iskounen, line 54: **suggestion:**

1. Can you use [Sass' `@extend` rule](https://sass-lang.com/documentation/at-rules/extend) to share these CSS rules with `.sidebar-link`?
2. Please use a more descriptive class name for this CSS rule.

app/views/admin/submissions/_collector_info.html.erb
iskounen: @Gon4ar5 I wasn't able to see how you addressed this from the latest commits to this PR. Would you mind providing a little explanation, please?

Gon4ar5: I added a new filter, and now it works like this: if a submission has a user, we display all submissions of this user, if not, then all submissions with the same user_email

iskounen (approved):

PR #1205, opened 2021-12-21
feat[SWA-76]: Add communication between OHM and Convection

Gon4ar5:
Resolve [SWA-76](https://artsyproduct.atlassian.net/browse/SWA-76)

In this pull I added fields that are updated after the end of the sale.

The main problem was that convection didn't always get artwork information, so I set the publish_artworks_access_granted gravity flag to true for the convection app.

### DIFF
--- a/app/controllers/api/consignments_controller.rb
+++ b/app/controllers/api/consignments_controller.rb
@@ -17,30 +17,20 @@ module Api
       sale_artworks =
         Gravity.client.sale_artworks(sale_id: sale_id).sale_artworks
 
-      sale_artworks.map do |sale_artwork|
+      sale_artworks.each do |sale_artwork|
         artwork = sale_artwork.artwork
-        submission = Submission.find_by(source_artwork_id: artwork.id)
+        submission =
+          Submission.with_source_artwork_id.find_by(
+            source_artwork_id: artwork.id
+          )
         next unless submission
 
-        submission.assign_attributes(
-          title: artwork.title,
-          medium: artwork.medium
+        SubmissionService.update_submission_info(artwork, submission)
+        PartnerSubmissionService.update_consignment_info(
+          sale,
+          sale_artwork,
+          submission
         )
-        submission.save!
-
-        consignment = submission.consigned_partner_submission
-        next unless consignment
-
-        price = sale_artwork.highest_bid.try(:amount_cents)
-        state = price ? 'sold' : 'bought in'
-        consignment.assign_attributes(
-          sale_price_cents: price || consignment.sale_price_cents,
-          sale_lot_number: sale_artwork.lot_number,
-          sale_date: sale.end_date,
-          state: state,
-          sale_name: sale.name
-        )
-        consignment.save!
       end
     end
   end
--- a/app/controllers/api/consignments_controller.rb
+++ b/app/controllers/api/consignments_controller.rb
@@ -13,37 +13,34 @@ module Api
     private
 
     def update_sale_info(sale_id)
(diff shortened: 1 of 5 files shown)


app/controllers/api/consignments_controller.rb
```diff
@@ -5,22 +5,42 @@ class ConsignmentsController < RestController
...
-    def fetch_sale_artworks_with_price(sale_id)
+    def update_sale_info(sale_id)
+      sale = Gravity.client.sale(id: sale_id)._get
       sale_artworks =
         Gravity.client.sale_artworks(sale_id: sale_id).sale_artworks
 
       sale_artworks.map do |sale_artwork|
-        artwork_id = sale_artwork.artwork.id
+        artwork = sale_artwork.artwork
+        submission = Submission.find_by(source_artwork_id: artwork.id)
```
iskounen, line 22: **suggestion (non-blocking):** We should consider indexing the `source_artwork_id` column depending on the performance of this endpoint. I've added a [ticket](https://artsyproduct.atlassian.net/browse/SWA-194) to the backlog for this work.

app/controllers/api/consignments_controller.rb
```diff
@@ -5,22 +5,42 @@ class ConsignmentsController < RestController
...
-        [artwork_id: artwork_id, price: price]
+        state = price ? 'sold' : 'bought in'
+        consignment.assign_attributes(
+          sale_price_cents: price || consignment.sale_price_cents,
+          sale_lot_number: sale_artwork.lot_number,
+          sale_date: sale.end_date,
+          state: state,
+          sale_name: sale.name
+        )
+        consignment.save!
```
iskounen, line 43: **suggestion:** We should update this logic in [`PartnerSubmissionService#update_price`](https://github.com/artsy/convection/blob/9918182bc6950f0442a1390d47128864e76beb26/app/services/partner_submission_service.rb#L108-L118) and call that method from here.

iskounen (changes requested):

app/controllers/api/consignments_controller.rb
```diff
@@ -5,22 +5,42 @@ class ConsignmentsController < RestController
...
       sale_artworks.map do |sale_artwork|
-        artwork_id = sale_artwork.artwork.id
+        artwork = sale_artwork.artwork
+        submission = Submission.find_by(source_artwork_id: artwork.id)
+        next unless submission
+
+        submission.assign_attributes(
+          title: artwork.title,
+          medium: artwork.medium
+        )
```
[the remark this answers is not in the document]
Gon4ar5: The ticket says that these submission fields need to be updated:
1. Artist/Designer
2. Title
3. Medium
4. Year
I update the first 3 (for the year I need to update gravity)

app/controllers/api/consignments_controller.rb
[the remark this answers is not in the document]
Gon4ar5: What do you think about creating the `with_source_artwork_id` scope?

Gon4ar5: It sounds good

app/services/submission_service.rb
```diff
@@ -276,5 +276,15 @@ def deliver_rejection_notification(submission_id)
         artist: artist
       ).deliver_now
     end
+
+    def update_submission_info(artwork, submission)
+      submission.assign_attributes(
+        title: artwork.title,
+        medium: artwork.medium,
+        year: artwork.date,
```
Gon4ar5, line 284: @iskounen 
Gravity already had a date field.

iskounen (approved):

PR #1217, opened 2022-01-17
[WIP] Display whether an artwork was submitted from web or mobile (SWA-168)

iskounen:
[SWA-168](https://artsyproduct.atlassian.net/browse/SWA-168)

This PR displays whether an artwork was submitted from the web or mobile submission flows. It uses the `user_agent` field that is provided when submissions are made.


iskounen:
Closing this pull request because we should use a more reliable and straightforward way to identify the source of submissions that includes Volt and Convection.

PR #1221, opened 2022-01-20
SWA-150: Create artwork and add it to my collection

Gon4ar5:
Resolve [SWA-150](https://artsyproduct.atlassian.net/browse/SWA-150)

Need to add description and fix specs

### DIFF
--- a/app/controllers/concerns/graphql_helper.rb
+++ b/app/controllers/concerns/graphql_helper.rb
@@ -30,7 +30,7 @@ module GraphqlHelper
   def create_my_collection_artwork(submission, x_app_token)
     Metaql::Schema.execute(
       query: my_collection_create_artwork_mutation_builder,
-      token: x_app_token,
+      x_app_token: x_app_token,
       variables: {
         input: my_collection_create_artwork_mutation_params(submission)
       }
@@ -104,9 +104,9 @@ module GraphqlHelper
       artistIds: [submission.artist_id],
       artworkLocation:
         [
-          submission.location_city,
-          submission.location_state,
-          submission.location_country
+          submission.location_city.to_s,
+          submission.location_state.to_s,
+          submission.location_country.to_s
         ].delete_if(&:empty?).join(', '),
       category: submission.category,
       date: submission.year,
--- a/app/controllers/concerns/graphql_helper.rb
+++ b/app/controllers/concerns/graphql_helper.rb
@@ -39,16 +39,12 @@ module GraphqlHelper
     GQL
   end
 
-  def create_my_collection_artwork(submission, current_user)
+  def create_my_collection_artwork(submission, _current_user)
     response =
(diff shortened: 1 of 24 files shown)


app/services/submission_service.rb
```diff
@@ -87,6 +89,10 @@ def update_submission(
           submission.assign_attributes(
             reject_non_target_supply_artist(submission.artist_id)
           )
+          if submission.submitted? &&
+               submission.user.save_submission_to_my_collection?
```
iskounen, line 93: **suggestion:** We should check that the submission was made by an authenticated user before creating an artwork in My Collection. For anonymous users, the artwork will be created once they create an Artsy account.

app/controllers/concerns/graphql_helper.rb
```diff
@@ -74,4 +127,31 @@ def match_partners_query(term)
 
     match_partners_response[:data][:match_partners]
   end
+
+  def my_collection_create_artwork_mutation_params(submission, current_user)
+    {
+      user_id: current_user,
```
iskounen, line 133: **suggestion:** We should use the `submission.user` ID instead of `current_user` in case we enable this feature for submissions created in Convection (or Volt) in the future.

app/services/submission_service.rb
Gon4ar5: if the user is not authorized, then `submission.user&.save_submission_to_my_collection?`  return nil

app/controllers/concerns/graphql_helper.rb
Gon4ar5: good catch, fixed

app/services/submission_service.rb
iskounen: Got it. Thanks for adding the safe navigation operator (`&.`) 🙌

iskounen: That's true for our current configuration. However, one day we will remove the feature flag and all submissions will be allowed to save submissions to My Collection. When that happens, we will need to remember that submissions from anonymous users should not be added to My Collection, so I think it's better if we prepare for that now.

Gon4ar5: Hmm, as far as I remember, if this is an anonymous submission, then it does not have a user, that is, in any case, submission.user == nil ==> submission.user&.save_submission_to_my_collection? == nil

iskounen: That is all correct, and will remain true until we launch the My Collection integration.

Once we launch the My Collection integration, we will remove the `swa_my_collection` feature flag as well as the `save_submission_to_my_collection?` method because we won't need them anymore.

When that happens, we will need to add an additional check for anonymous users. What I'm proposing is that we add the check for anonymous users today, so that we don't forget to add that when we remove the feature flag in the future.

iskounen (approved):

PR #1222, opened 2022-01-24
[WIP]fix(SWA-206): create new User when an Admin creates a submission

kiryl-zubarau:
This PR creates a new user record for the admin on submission creation.

JIRA - https://artsyproduct.atlassian.net/browse/SWA-206
Slack - https://artsy.slack.com/archives/C02B4Q2TKGW/p1642794932077500

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -29,7 +29,8 @@ class SubmissionService
 
       if AdminUser.exists?(gravity_user_id: current_user)
         create_params.merge!(
-          created_by: User.find_by(gravity_user_id: current_user).email
+          created_by:
+            User.find_or_create_by(gravity_user_id: current_user).email
         )
       end
 


app/services/submission_service.rb
```diff
@@ -29,7 +29,8 @@ def create_submission(
 
       if AdminUser.exists?(gravity_user_id: current_user)
         create_params.merge!(
-          created_by: User.find_by(gravity_user_id: current_user).email
+          created_by:
+            User.find_or_create_by(gravity_user_id: current_user).email
```
kiryl-zubarau, line 33: Also, I suppose it makes sense to update the logic of `AdminUser`creation. For now, it just creates a record at `AdminUser` table so I think we can create a new `User` here.

https://github.com/artsy/convection/blob/d28c622c9e46e60be0dc3eaa975a04df49431452/app/controllers/admin/admin_users_controller.rb#L30-L37

What do you all think about that?

iskounen: I think that creating a new `User` record when an `AdminUser` is created, or when an `AdminUser` creates a submission should be avoided. There's no reason for these two models to be linked because they represent different people: a `User` is a collector who is selling their artwork, and an `AdminUser` is an Artsy employee who assists in the selling process.

While your change in this PR will solve the bug that admins are experience, I think that it will deepen the link between the `User` and `AdminUser` models, and ultimately make Convection more complex.

I propose that we either:

1. Save the admin's `gravity_user_id` in the `created_by` field instead of their email, or
2. Fetch the admin's email from Gravity when creating a submission, or
3. Add admin emails to the `AdminUser` model

(1) is probably the easiest option that lets us separate the `User` and `AdminUser` models.

kiryl-zubarau: I would prefer 3rd variant, but do it automatically, load user information from `Gravity` when `AdminUser` creating and save the email to the `AdminUser`, then we can use the email the `created_by` field. I think it will be more useful for the admins to see email instead of guid.

kiryl-zubarau: Also, I can write a migration script for the current admins to prefill emails.

iskounen: I agree that (3) is the best option, and appreciate you writing a migration script for the existing admins (there are about 25, of which about half don't have `User` records in Convection.)

If we go with option (3), what do you think about the idea of changing the `created_by` field to be a foreign key to the `AdminUser` record instead of an email address? The benefit is that we can update the email address (or other details) of the admin and still have it reflected accurately on past submissions.

kiryl-zubarau: Sounds good to me!

PR #1223, opened 2022-01-25
[WIP] [SWA-208]: Creates external_id column in submissions table

nickskalkin:
Creates external_id column in submissions table.

Jira card - https://artsyproduct.atlassian.net/jira/software/c/projects/SWA/boards/113?modal=detail&selectedIssue=SWA-208

### DIFF
--- a/db/migrate/20220125100834_add_external_id_to_submissions.rb
+++ b/db/migrate/20220125100834_add_external_id_to_submissions.rb
@@ -1,0 +1,12 @@
+class AddExternalIdToSubmissions < ActiveRecord::Migration[6.1]
+  def change
+    enable_extension 'pgcrypto' unless extension_enabled?('pgcrypto') # need to enable pgcrypto to use `gen_random_uuid()` function
+
+    add_column :submissions,
+               :external_id,
+               :uuid,
+               default: 'gen_random_uuid()',
+               null: false
+    add_index :submissions, :external_id, unique: true
+  end
+end
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1664,6 +1664,16 @@ enum SubmissionSort {
   EXHIBITION_DESC
 
   """
+  sort by external_id in ascending order
+  """
+  EXTERNAL_ID_ASC
+
+  """
+  sort by external_id in descending order
+  """
+  EXTERNAL_ID_DESC
+
+  """
   sort by ext_user_id in ascending order
   """
   EXT_USER_ID_ASC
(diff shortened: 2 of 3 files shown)


db/migrate/20220125100834_add_external_id_to_submissions.rb
```diff
@@ -0,0 +1,12 @@
+class AddExternalIdToSubmissions < ActiveRecord::Migration[6.1]
+  def change
+    enable_extension 'pgcrypto' unless extension_enabled?('pgcrypto') # need to enable pgcrypto to use `gen_random_uuid()` function
```
nickskalkin, line 3: Had to enable `pgcrypto`, otherwise `gen_random_uuid()` wouldn't work

Gon4ar5 (approved):
LGTM!

kiryl-zubarau (approved):

db/migrate/20220125100834_add_external_id_to_submissions.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    add_column :submissions,
+               :external_id,
```
iskounen, line 6: **question:** What will `external_id` represent? My understanding is that:

- `id` will continue being the sequential, numerical ID
- `uuid` will be the non-sequential, generated ID

I assumed that we would expose the `uuid` field to clients as an "external" ID so I am wondering why we have a separate `external_id` column.

nickskalkin: @iskounen `external_id` is basically a `uuid` column you are talking about. Clients will be working with submissions using this column. Do you propose to rename `external_id` to `uuid`? I actually like `uuid` more

iskounen: Thanks for the explanation. I was reading the migration file incorrectly... 😅 

Regarding the name, my preference is to use `uuid` on the data model, and decide how we want to expose that to clients at the GraphQL layer.

nickskalkin: I will move this PR to WIP and rename db column. @kiryl-zubarau @iskounen what is your opinion on how to name this field in API?

kiryl-zubarau: I would prefer `external id`

nickskalkin:
I am closing this PR in favour of https://github.com/artsy/convection/pull/1224

The second one is based on the current one. I had to make changes in the migration and I decided to make in the second PR,  just because it is easier. Let me know if you still prefer that I make changes in this PR and align history of the second one

cc @kiryl-zubarau @iskounen

PR #1224, opened 2022-01-26
SWA-210: GraphQL support for Submission#external_id

nickskalkin:
Jira card - https://artsyproduct.atlassian.net/jira/software/c/projects/SWA/boards/113?modal=detail&selectedIssue=SWA-210

### DIFF
--- a/app/graphql/resolvers/add_asset_to_submission_resolver.rb
+++ b/app/graphql/resolvers/add_asset_to_submission_resolver.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 class AddAssetToSubmissionResolver < BaseResolver
-  include SubmissionableResolver
+  include Resolvers::Submissionable
 
   def run
-    raise(GraphQL::ExecutionError, 'Submission Not Found') unless submission
+    check_submission_presence!
 
     unless matching_user(submission, @arguments&.[](:session_id)) || admin?
       raise(GraphQL::ExecutionError, 'Submission Not Found')
@@ -12,7 +12,7 @@ class AddAssetToSubmissionResolver < BaseResolver
 
     @arguments[:asset_type] ||= 'image'
 
-    asset = submission.assets.create!(@arguments.except(:session_id))
+    asset = submission.assets.create!(asset_params)
     SubmissionService.notify_user(submission.id) if submission.submitted?
 
     { asset: asset }
@@ -20,11 +20,17 @@ class AddAssetToSubmissionResolver < BaseResolver
 
   private
 
+  # overwrites Resolvers::Submissionable
   def submission_id
     @arguments[:submission_id]
   end
 
+  # overwrites Resolvers::Submissionable
   def external_submission_id
     @arguments[:external_submission_id]
+  end
+
+  def asset_params
+    @arguments.except(:session_id, :external_submission_id)
   end
 end
--- a/app/graphql/resolvers/add_asset_to_submission_resolver.rb
+++ b/app/graphql/resolvers/add_asset_to_submission_resolver.rb
@@ -1,15 +1,10 @@
 # frozen_string_literal: true
 
 class AddAssetToSubmissionResolver < BaseResolver
-  def valid?
-    true
-  end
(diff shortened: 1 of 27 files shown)


app/graphql/resolvers/add_asset_to_submission_resolver.rb
```diff
@@ -1,25 +1,36 @@
 # frozen_string_literal: true
 
 class AddAssetToSubmissionResolver < BaseResolver
-  def valid?
```
nickskalkin, line 4: `valid?` method now is present in `include Resolvers::Submissionable`

app/graphql/resolvers/add_asset_to_submission_resolver.rb
```diff
@@ -1,25 +1,36 @@
...
-    true
-  end
+  include Resolvers::Submissionable
 
   def run
-    submission = Submission.find_by(id: @arguments[:submission_id])
-    unless submission
-      raise(GraphQL::ExecutionError, 'Submission from ID Not Found')
-    end
+    check_submission_presence!
```
nickskalkin, line 7: I extracted lookup by submission id / external uuid to a separate module (Resolvers::Submissionable) to avoid code duplication.

app/graphql/resolvers/add_asset_to_submission_resolver.rb
```diff
@@ -1,25 +1,36 @@
...
-    asset = submission.assets.create!(@arguments.except(:session_id))
+    asset = submission.assets.create!(asset_params)
     SubmissionService.notify_user(submission.id) if submission.submitted?
 
     { asset: asset }
   end
+
+  private
+
+  # overwrites Resolvers::Submissionable
```
nickskalkin, line 23: if it's not clear what's going on here - please see docs in Resolvers::Submissionable. If still not clear - let's discuss how to implement it better

app/graphql/resolvers/concerns/resolvers/submissionable.rb
```diff
@@ -0,0 +1,45 @@
...
+# `external_submission_id` in your resolvers.
+
+module Resolvers::Submissionable
+  IdsNotPassed =
+    GraphQL::ExecutionError.new('Neither id nor externalId have been passed.')
+
+  def submission
+    @submission ||=
+      Submission.find_by(id: submission_id) ||
+        Submission.find_by(uuid: external_submission_id)
```
kiryl-zubarau, line 16: **non-blocking**: maybe it makes sense to change order and look for submissions by UUID first.

app/services/submission_service.rb
```diff
@@ -39,7 +39,7 @@ def create_submission(
         )
       end
 
-      submission = Submission.create!(create_params)
+      submission = Submission.create!(create_params).reload
```
kiryl-zubarau, line 42: **non-blocking**: let's add a comment here 😄  to show other people why we added `reload`

kiryl-zubarau (approved):
LGTM

app/graphql/resolvers/concerns/resolvers/submissionable.rb
kiryl-zubarau: Hmm, just realized it might be a blocker since UUID will be cast to a number.

nickskalkin: @kiryl-zubarau in this case there will be no such problem, since GraphQL clients differentiate `submission_id` and `external_submission_id` on their side. So it shouldn't be the case when force sends a query or mutation with `submission(id: "<uuid>")`

kiryl-zubarau: Missed that, thanks!

app/services/submission_service.rb
nickskalkin: @kiryl-zubarau please check my last PR. I've created a module which is responsible for automatic reload of uuid field

Gon4ar5 (approved):
Great work! LGTM

PR #1225, opened 2022-01-28
fix mailer previews that are not rendering

iskounen:
A majority of the mailer previews (`/rails/mailers`) were not rendering because of `user_detail`  fields that were added to fixtures during the work to enable anonymous submissions. This PR does two things:

1. Removes `user_detail` from fixtures that don't need it.
2. Adds the various rejection email templates to previews.


Gon4ar5 (approved):
LGTM

PR #1227, opened 2022-01-31
fix(SWA-206): add email field for admin users

kiryl-zubarau:
This PR adds an email field for admin users.

JIRA - https://artsyproduct.atlassian.net/browse/SWA-206

PR #1228, opened 2022-01-31
fix(SWA-206): add foreign key from submission to admin user

kiryl-zubarau:
~~**Do not merge the PR! Requires https://github.com/artsy/convection/pull/1227 to be deployed and executed a rake task**~~

This PR adds foreign key from submission to admin user and updates related logic.

JIRA - https://artsyproduct.atlassian.net/browse/SWA-206

### DIFF
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -77,7 +77,8 @@ module Admin
       begin
         user =
           Gravity.client.user(id: @admin_user.gravity_user_id).user_detail._get
-        @admin_user.email = user.email unless user.nil?
+
+        @admin_user.email = user.email if user
       rescue Faraday::ResourceNotFound
         nil
       end
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -28,6 +28,7 @@ module Admin
     # POST /admin/admin_users
     def create
       @admin_user = AdminUser.new(admin_user_params)
+      load_admin_email
 
       if @admin_user.save
         redirect_to admin_admin_users_path,
@@ -70,6 +71,16 @@ module Admin
       return if super_admin_user? @current_user
 
       raise ApplicationController::NotAuthorized
+    end
+
+    def load_admin_email
+      begin
+        user =
+          Gravity.client.user(id: @admin_user.gravity_user_id).user_detail._get
+        @admin_user.email = user.email unless user.nil?
+      rescue Faraday::ResourceNotFound
+        nil
+      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/controllers/admin/admin_users_controller.rb
```diff
@@ -71,5 +72,15 @@ def authorize_user!
 
       raise ApplicationController::NotAuthorized
     end
+
+    def load_admin_email
+      begin
+        user =
+          Gravity.client.user(id: @admin_user.gravity_user_id).user_detail._get
```
iskounen, line 79: **question (non-blocking):** Is there a Metaphysics endpoint that we can use instead of directly requesting user information from Gravity? Ideally all requests between Artsy services go through Metaphysics, and I'd like to see if we can start following that pattern in Convection.

db/migrate/20220128111006_add_foreign_key_from_submission_to_admin_user.rb
```diff
@@ -0,0 +1,41 @@
+class AddForeignKeyFromSubmissionToAdminUser < ActiveRecord::Migration[6.1]
+  def up
+    add_column :submissions, :created_by_id, :bigint
```
iskounen, line 3: **suggestion:** We may want to take this opportunity to rename this column so that its intention is clear. At the moment it's not clear why some submissions have a `created_by` and some don't, because all submissions were created by _someone_.

What do you think about something simple like `admin_id`?

app/controllers/admin/admin_users_controller.rb
kiryl-zubarau: We added a Metaphysics client at the [PR](https://github.com/artsy/convection/pull/1221) but it's not merged. For now, Convection doesn't use Metaphysics to load information as far as I know.

db/migrate/20220128111006_add_foreign_key_from_submission_to_admin_user.rb
kiryl-zubarau: Updated

db/migrate/20220128111006_add_foreign_key_from_submission_to_admin_user.rb
```diff
@@ -0,0 +1,41 @@
+class AddForeignKeyFromSubmissionToAdminUser < ActiveRecord::Migration[6.1]
+  def up
+    add_column :submissions, :admin_id, :bigint
+    add_foreign_key :submissions, :admin_users, column: :admin_id
+
+    populate_admin_id_field
```
iskounen, line 6: ❤️

iskounen (approved):

app/controllers/admin/admin_users_controller.rb
```diff
@@ -71,5 +72,15 @@ def authorize_user!
...
+        @admin_user.email = user.email unless user.nil?
```
nickskalkin, line 80: Just a suggestion: you can omit `.nil?` check here probably

PR #1232, opened 2022-02-01
Fix: Timezone specs

Gon4ar5:
I opened this PR due to timezone issues in tests, because my timezone is +3, the email send time becomes 26 Sept 21:00, while the expected date is 27 Sept, so I changed the send time to 12 hours of the day.
Screen Shot 2022-02-01 at 6 28 36 PM

### DIFF
--- a/spec/services/partner_submission_service_spec.rb
+++ b/spec/services/partner_submission_service_spec.rb
@@ -24,7 +24,7 @@ describe PartnerSubmissionService do
     stub_gravity_partner_communications
     stub_gravity_partner_contacts
     allow(Time).to receive(:now).and_return(
-      Time.new(2_017, 9, 27).in_time_zone('UTC')
+      Time.new(2_017, 9, 27, 12).in_time_zone('UTC')
     ) # stub time for email subject lines
     allow(Convection.config).to receive(:auction_offer_form_url).and_return(
       'https://google.com/auction'


spec/services/partner_submission_service_spec.rb
```diff
@@ -24,7 +24,7 @@
     stub_gravity_partner_communications
     stub_gravity_partner_contacts
     allow(Time).to receive(:now).and_return(
-      Time.new(2_017, 9, 27).in_time_zone('UTC')
+      Time.new(2_017, 9, 27, 12).in_time_zone('UTC')
```
iskounen, line 27: **question (non-blocking):** Will an engineer working in the [UTC+13:00](https://en.wikipedia.org/wiki/UTC%2B13:00) or [UTC:14:00](https://en.wikipedia.org/wiki/UTC%2B14:00) timezones encounter the same test failures that you experienced?

iskounen (approved):
Thanks for opening this PR. I think it exposes some assumptions that our test suite makes about digest emails that would be nice to fix, but I think this is good for now.

spec/services/partner_submission_service_spec.rb
Gon4ar5: The problem will be solved for those who are in the +12 time zone (they will have 27 Sept 00:00 in the test), but not for those who are in +13. And it will appear for those who are in the -12 time zone (but I think the number of such people is less than in the [+1...+12] time zone)

PR #1238, opened 2022-02-07
Fix[SWA-150]: Add submissionId field to artwork creation mutation

Gon4ar5:
Fix for [SWA-150](https://artsyproduct.atlassian.net/browse/SWA-150)
Added submissionID field to mutation, small fixes


kiryl-zubarau (approved):

iskounen (approved):
I'm merging this to test the access token approach on staging. (The parts that make use of the access token and affect Gravity are behind a feature flag, so I feel safe testing this on staging.)

PR #1239, opened 2022-02-09
[SWA-226] Fix detailed rejection emails with rejection reason are not being sent

fladson:
### What is the problem?
Convection has several templates for rejection emails that each correspond to a reason for which a submission can be rejected (artist submission, fake artwork, etc.)

In the current manual rejection flow, an email is sent before an admin has a chance to select a rejection reason. As a result, collectors receive the default rejection email template, which does not explain why their submission was rejected.

### Why does this matter?
One of our quarterly goals is to improve the NPS (net promoter score) of the Sell with Artsy flow, and user research shows that the shock of being rejected is a contributing factor to that low score.

### What is the solution?
When an admin manually rejects a submission, do not immediately schedule the email to be sent.

Send a rejection email when the rejection reason is updated on a submission instead.

### What should happen if an admin never selects a rejection reason?
We will update the rejection flow in Convection to show the rejection reasons after “Rejection” is clicked, and only send an email after one of the reasons is selected.

If an admin later updates the rejection reason after the initial rejection, we won’t send an email to the user, but we will update the submission in the database.
<hr/>

The solution we found after some iterations is to replace the current rejection flow with a modal. The modal will prevent us from having a rejected submission without a rejection reason.

https://user-images.githubusercontent.com/554507/153185815-d0510b16-0ec0-4be9-abe0-4867e357f4b2.mov

### DIFF
--- a/app/assets/stylesheets/locals.css.scss
+++ b/app/assets/stylesheets/locals.css.scss
@@ -239,13 +239,19 @@ a.make-primary-link {
   pointer-events: none;
 }
 
-.offer-type-radio {
+.offer-type-radio,
+.submission-reason-radio{
   display: flex;
   justify-content: flex-start;
   align-items: center;
 
   input {
     width: 30px;
+  }
+
+  label {
+    margin-top: 10px;
+    margin-left: 5px;
   }
 }
 
--- a/app/assets/stylesheets/locals.css.scss
+++ b/app/assets/stylesheets/locals.css.scss
@@ -334,3 +334,9 @@ a.make-primary-link {
 .align-center {
   text-align: center;
 }
+
+@media (min-width: $screen-sm-min) {
+  .smaller-modal {
+    height: 500px !important;
+  }
+}
--- a/app/views/admin/submissions/_rejection_reasons_modal.html.erb
+++ b/app/views/admin/submissions/_rejection_reasons_modal.html.erb
@@ -16,7 +16,7 @@
           <div class='row'>
             <div class='col-md-12'>
               <% Submission::REJECTION_REASONS.each do |rejection_reason| %>
-                <div class='offer-type-radio'>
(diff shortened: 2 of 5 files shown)


kiryl-zubarau:
Looks good! Just a few UI notes: 
1. Can you make popup height the same as content size, I would like to get rid of white space under the save button
2. Add space between a radio button and a reason label.
3. Align a reason label in the center (vertically)

fladson:
> Looks good! Just a few UI notes:
> 
> 1. Can you make popup height the same as content size, I would like to get rid of white space under the save button
> 2. Add space between a radio button and a reason label.
> 3. Align a reason label in the center (vertically)
> 4. Disable the save button until you select a reason

Thanks for the review, initially I just copy-pasted the same modal we have for offer rejections and didn't want to change much but I pushed some changes to improve both.

_reason selection is now required_
I opted to have this instead of a logic to control the button state, let me know what do you think.
Screenshot 2022-02-09 at 17 43 49

_vertical alignment adjustments_
Screenshot 2022-02-09 at 17 43 24
Screenshot 2022-02-09 at 17 46 16

Gon4ar5 (approved):
LGTM!
But I think you can remove 'Please select one'
Screen Shot 2022-02-10 at 4 24 11 PM

https://github.com/artsy/convection/blob/main/app/views/admin/submissions/show.html.erb#L67

fladson:
> LGTM! But I think you can remove 'Please select one' Screen Shot 2022-02-10 at 4 24 11 PM
> 
> https://github.com/artsy/convection/blob/main/app/views/admin/submissions/show.html.erb#L67

Nice catch, thanks! [b4a9600](https://github.com/artsy/convection/pull/1239/commits/b4a9600f6170aaeb46be1d35402b76f6047a3bcb)

PR #1254, opened 2022-02-24
Fix[SWA-262]: My collection artwork is not created if submission state is rejected

Gon4ar5:
Resolve [SWA-262](https://artsyproduct.atlassian.net/browse/SWA-262)

My collection artwork must be created after the user is done with it, no matter what state it is in after the draft state.

Also added a check that if the artwork has already been created, then the next time the submission is updated, we will not create a second artwork

Add tests

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -90,7 +90,8 @@ class SubmissionService
           submission.assign_attributes(
             reject_non_target_supply_artist(submission.artist_id)
           )
-          if submission.submitted? && submission.user && !access_token.nil? &&
+          if !submission.draft? && submission.user && !access_token.nil? &&
+               !submission.my_collection_artwork_id &&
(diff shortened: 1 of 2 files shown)


fladson (approved):
LGTM, thanks for adding the specs, I also suffered a bit yesterday to cover that condition.

This method/condition is only growing in complexity, what do you think of extracting this logic to a different method or class (in a separate ticket of course)?

Gon4ar5:
> LGTM, thanks for adding the specs, I also suffered a bit yesterday to cover that condition.
> 
> This method/condition is only growing in complexity, what do you think of extracting this logic to a different method or class (in a separate ticket of course)?

Yes, I think we really should refactor the `create_submission` and `update_submission` methods, cause it's already difficult to understand what is happening in them

kiryl-zubarau (approved):

PR #1255, opened 2022-02-24
Ensure trusted apps have that role

jonallured:
Unless I'm missing something I believe we are not checking for the `trusted` role for the GQL api on convection when the request comes from a client application. These two tests demonstrate my concerns.

### DIFF
--- a/app/graphql/resolvers/base_resolver.rb
+++ b/app/graphql/resolvers/base_resolver.rb
@@ -21,7 +21,7 @@ class BaseResolver
   private
 
   def trusted_application?
-    @context[:current_application].present?
+    @context[:current_application].present? && @context[:current_user_roles].include?(:trusted)
   end
 
   def user?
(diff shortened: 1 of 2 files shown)


jonallured:
Lot's more chatter here: https://artsy.slack.com/archives/CQJMKMZ6X/p1645718812692369

joeyAghion (approved):
This looks totally reasonable to me.

I see that `trusted_application?` is only used [here](https://github.com/artsy/convection/blob/2aa1669f46199e442ac2d9ce92e5c783e212fe23/app/graphql/resolvers/create_offer_resolver.rb#L5). Do all the other graphql operations do adequate authorization already?

jonallured:
> Do all the other graphql operations do adequate authorization already?

Valid question! I don't feel comfortable making a call on that - maybe this is something I should follow up on with @iskounen since his team owns convection now?

PR #1256, opened 2022-02-24
[SWA-212] Trigger braze email campaign after a user completes a submission

fladson:
Note: _**This is a POC code, it should not be used in production. We don't have copies of the email subject and body yet.**_

This should cover the first case:

When an authenticated user submits an artwork
- and they have opted-in to receive marketing emails
  - and they already have the app, then they **receive** an email 24 hours after the submission was submitted that reminds them to see their artwork in My Collection.
  - and they do not have the app, then they **receive** an email 24 hours after the submission was submitted that reminds them to download the app and see their artwork in My Collection.
- and they have not opted-in to receive marketing emails, then they **never** receive an email that reminds them to download the app or see their artwork in My Collection.

TODO:

- [ ] Test on staging
  - [ ] Test a case where Braze doesn't yet have the user in the DB
- [ ] Upload the new `.env.shared` on both environments

### DIFF
--- a/app/services/braze_service.rb
+++ b/app/services/braze_service.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class BrazeService
+  class << self
+    def send_swa_my_collection_email(submission_id)
+      submission = Submission.find(submission_id)
+      raise 'Still processing images.' unless submission.ready?
+
+      raise 'User lacks email.' if submission.email.blank?
+
+      recipient = [
+        {
+          external_user_id: submission.user.gravity_user_id,
+          trigger_properties: {
+            email_subject: '[TESTING] SWA Braze integration'
+          }
+        }
+      ]
+      braze_campaign_id = Convection.config.braze_campaign_id
+
+      BrazeApi.trigger_campaign_send(braze_campaign_id, recipient)
+    end
+  end
+end
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -67,6 +67,7 @@ class SubmissionService
       params
     end
 
+    # This method deserves a refactor, there are a lot of things happening here
     def update_submission(
       submission,
       params,
@@ -92,13 +93,20 @@ class SubmissionService
           )
           if submission.submitted? && submission.user && !access_token.nil? &&
                submission.user&.save_submission_to_my_collection?
-            create_my_collection_artwork(submission, access_token)
+            response = create_my_collection_artwork(submission, access_token)
+            notify_user_with_braze(submission.id, response)
           end
         end
 
         update_submission_state(submission, current_user)
       end
       submission.save!
+    end
+
+    def notify_user_with_braze(submission_id, response)
+      if response
+        SwaMyCollectionEmailWorker.perform_in(1.day.from_now, submission_id)
+      end
     end
 
     def update_submission_state(submission, current_user)
--- a/spec/services/braze_service_spec.rb
+++ b/spec/services/braze_service_spec.rb
@@ -31,7 +31,7 @@ describe BrazeService do
     context 'when submission is valid' do
       let(:gravity_user_id) { 'userid' }
       let(:email_subject) { '[TESTING] SWA Braze integration' }
-      let(:braze_campaign_id) { '244d6288-4f71-4c49-a6fb-9df00826550a' }
+      let(:braze_campaign_id) { Convection.config.braze_campaign_id }
       let(:submission) do
         Fabricate(
           :submission,
--- a/spec/services/braze_service_spec.rb
+++ b/spec/services/braze_service_spec.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe BrazeService do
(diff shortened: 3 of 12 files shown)


spec/services/braze_service_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        expect {
+          BrazeService.send_swa_my_collection_email(submission.id)
+        }.to raise_error('User lacks email.')
+      end
+    end
+
+    context 'when submission is valid' do
+      let(:gravity_user_id) { 'userid' }
+      let(:email_subject) { '[TESTING] SWA Braze integration' }
+      let(:braze_campaign_id) { '244d6288-4f71-4c49-a6fb-9df00826550a' }
```
fladson: This got under my radar and I already deactivated the mentioned campaign.

app/services/braze_service.rb
```diff
@@ -0,0 +1,24 @@
...
+      submission = Submission.find(submission_id)
+      raise 'Still processing images.' unless submission.ready?
+
+      raise 'User lacks email.' if submission.email.blank?
+
+      recipient = [
+        {
+          external_user_id: submission.user.gravity_user_id,
+          trigger_properties: {
+            email_subject: '[TESTING] SWA Braze integration'
```
fladson, line 15: This needs a final copy.

fladson: We also need the template ready to be able to send other fields here.

app/services/submission_service.rb
```diff
@@ -101,6 +103,12 @@ def update_submission(
       submission.save!
     end
 
+    def notify_user_with_braze(submission_id, response)
+      if response
+        SwaMyCollectionEmailWorker.perform_in(1.day.from_now, submission_id)
+      end
+    end
+
```
fladson, line 111: This is here only because Rubocop was raising that there are 2 levels of conditions on that method. We need a separate ticket to refactor this whole class.

app/services/submission_service.rb
```diff
@@ -101,6 +103,12 @@ def update_submission(
       submission.save!
     end
 
+    def notify_user_with_braze(submission_id, response)
+      if response
+        SwaMyCollectionEmailWorker.perform_in(1.day.from_now, submission_id)
```
fladson, line 108: I decided to use a worker instead of the `delay_until` just to have a better overview of failures and such within sidekiq.

app/services/braze_service.rb
```diff
@@ -0,0 +1,24 @@
...
+  class << self
+    def send_swa_my_collection_email(submission_id)
+      submission = Submission.find(submission_id)
+      raise 'Still processing images.' unless submission.ready?
+
+      raise 'User lacks email.' if submission.email.blank?
+
+      recipient = [
+        {
+          external_user_id: submission.user.gravity_user_id,
```
fladson, line 13: We need to play more with `external_user_id` and `user_alias`, I think the gem is not supporting sending only the `user_alias`, so I believe this would support only cases where the user is already synched in Braze.

spec/services/submission_service_spec.rb
```diff
@@ -479,6 +480,50 @@
       expect(submission.rejection_reason).to eq nil
     end
 
+    context 'Braze integration' do
```
fladson, line 483: If #1254 is merged before, we need to update this.

kiryl-zubarau (approved):
LGTM!

spec/services/submission_service_spec.rb
kiryl-zubarau: It was merged a few minutes ago 🙁

app/services/braze_service.rb
iskounen: Gravity keeps user accounts synched with Braze, so it is safe to assume that if a submission has a `gravity_user_id` that it is synched to Braze.

app/services/braze_service.rb
iskounen: I think we can use the existing `swa_my_collection` feature flag if you would like to try triggering an email.

app/services/submission_service.rb
```diff
@@ -67,6 +67,7 @@ def reject_non_target_supply_artist(artist_id)
       params
     end
 
+    # This method deserves a refactor, there are a lot of things happening here
```
iskounen, line 70: ❤️

app/services/submission_service.rb
iskounen: That's a great idea! Do we have the option to re-send any failed attempts with a worker?

app/services/submission_service.rb
iskounen: Here is an empty ticket if you'd like to add any details to guide the refactoring:

https://artsyproduct.atlassian.net/browse/SWA-265

fladson:
_From @iskounen_ 
> We initially built this out in Convection, but realized that it made more sense to trigger the Braze notification (email or IAM) from Gravity when an artwork is added to a user’s collection (with a submission ID and user ID).

Closing this ticket in favor of [SWA-269](https://artsyproduct.atlassian.net/browse/SWA-269)

PR #1259, opened 2022-03-01
feat(SWA-61)Add restricted field and update submissions query to work with it

kiryl-zubarau:
This PR allows users to load the information about their own submissions via GraphQL.

JIRA - https://artsyproduct.atlassian.net/browse/SWA-261

<details><summary>User loads own submission</summary>

image

</details>

<details><summary>User loads someone else's submission</summary>

image

</details>

<details><summary>Admin loads submission</summary>

image

</details>

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1303,7 +1303,6 @@ type Submission {
   attributionClass: AttributionClass
   authenticityCertificate: Boolean
   category: String
-  consignmentState: String
   createdAt: ISO8601DateTime
   currency: String
   depth: String
@@ -1333,6 +1332,7 @@ type Submission {
   provenance: String
   publishedAt: ISO8601DateTime
   rejectionReason: String
+  saleState: String
   signature: Boolean
 
   """
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1303,6 +1303,7 @@ type Submission {
   attributionClass: AttributionClass
   authenticityCertificate: Boolean
   category: String
+  consignmentState: String
   createdAt: ISO8601DateTime
   currency: String
   depth: String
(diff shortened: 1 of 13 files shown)


app/graphql/resolvers/submissions_resolver.rb
```diff
@@ -21,14 +22,22 @@ def valid?
...
 
   def run
-    base_submissions.where(conditions).order(sort_order)
+    submissions = base_submissions.where(conditions).order(sort_order)
+
+    submissions.map do |submission|
+      if admin? || partner? || submission.draft?
+        submission
+      else
+        submission.as_json(properties: :short)
```
fladson, line 31: We don't need the `.order(sort_order)` anymore?

kiryl-zubarau: Do you mean that the order of the submissions will be broken?

fladson: I said that based on the diff.
`base_submissions.where(conditions).order(sort_order)`

_schema.graphql
```diff
@@ -1303,6 +1303,7 @@ type Submission {
   attributionClass: AttributionClass
   authenticityCertificate: Boolean
   category: String
+  consignmentState: String
```
AlicanAkyuz, line 1306: **question/suggestion**: after removing the references to the word consignment, I think we are trying to avoid using it in our codebase as much as possible. do you think we can rename this to `submissionState` or `submissionStatus`?

kiryl-zubarau: Updated to `sale_state`

iskounen (approved):

PR #1264, opened 2022-03-14
feat(SWA-235): add submission source

kiryl-zubarau:
The PR adds source field to identify the source of the submission.

JIRA - https://artsyproduct.atlassian.net/browse/SWA-235

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -35,6 +35,8 @@ class SubmissionService
         )
       end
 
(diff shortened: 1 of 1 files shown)


fladson (approved):
LGTM, can we add a spec covering this case?

iskounen:
I didn't see @fladson's message before I merged, but I think that's a great idea. @kiryl-zubarau could you open a PR to test that, please?

PR #1265, opened 2022-03-14
feat: Add sourceArtworkID to UpdateSubmissionMutation

olerichter00:
Addresses [CX-2384]

## Description

Add sourceArtworkID to UpdateSubmissionMutation.

[CX-2384]: https://artsyproduct.atlassian.net/browse/CX-2384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/graphql/mutations/update_submission_mutation.rb
+++ b/app/graphql/mutations/update_submission_mutation.rb
@@ -32,6 +32,10 @@ module Mutations
     argument :provenance, String, required: false
     argument :signature, Boolean, required: false
     argument :state, Types::StateType, required: false
+    argument :sourceArtworkID,
+             String,
+             required: false,
+             description: 'If this artwork exists in Gravity, its ID'
     argument :title, String, required: false
     argument :width, String, required: false
     argument :year, String, required: false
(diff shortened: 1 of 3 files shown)


app/graphql/mutations/update_submission_mutation.rb
```diff
@@ -32,6 +32,10 @@ class UpdateSubmissionMutation < Mutations::BaseMutation
     argument :provenance, String, required: false
     argument :signature, Boolean, required: false
     argument :state, Types::StateType, required: false
+    argument :sourceArtworkID,
+             String,
+             required: false,
+             description: 'If this artwork exists in Gravity, its ID'
```
kiryl-zubarau, line 38: Why do you want to add the field to the UpdateSubmissionMutation? I suppose it should be enough to just initialize the value on submission creation. It's actual also for `my_collection_artwork_id`

kiryl-zubarau: Have you tried to move `sourceArtworkID` directly to the submission create call as it happens for the `id` field on submission update?
image

olerichter00: This works 🌟 I'll close this PR and go with your solution.

PR #1269, opened 2022-03-16
[dev]: add new env var to the config

AlicanAkyuz:
### Description 

This PR adds our new ENV variable to the config file. This new ENV is set to true on staging and false on production, and it will be utilised to determine whether or not we send a new version of receipt email (which'll encourage anonymous visitor to create an account with Artsy) after a submission is completed.

### DIFF
--- a/config/initializers/_config.rb
+++ b/config/initializers/_config.rb
@@ -53,5 +53,5 @@ Convection.config =
       ENV['VIBRATIONS_URL'] || 'https://admin-partners-staging.artsy.net',
     unleash_url: ENV['UNLEASH_URL'],
     unleash_token: ENV['UNLEASH_TOKEN'],
-    send_new_receipt_email: ENV['SEND_NEW_RECEIPT_EMAIL'] || 'false'
+    send_new_receipt_email: ENV['SEND_NEW_RECEIPT_EMAIL'] == 'true'
   )
--- a/config/initializers/_config.rb
+++ b/config/initializers/_config.rb
@@ -53,5 +53,5 @@ Convection.config =
       ENV['VIBRATIONS_URL'] || 'https://admin-partners-staging.artsy.net',
     unleash_url: ENV['UNLEASH_URL'],
     unleash_token: ENV['UNLEASH_TOKEN'],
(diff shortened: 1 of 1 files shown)


config/initializers/_config.rb
```diff
@@ -52,5 +52,6 @@ module Convection
     vibrations_url:
       ENV['VIBRATIONS_URL'] || 'https://admin-partners-staging.artsy.net',
     unleash_url: ENV['UNLEASH_URL'],
-    unleash_token: ENV['UNLEASH_TOKEN']
+    unleash_token: ENV['UNLEASH_TOKEN'],
+    send_new_receipt_email: ENV['SEND_NEW_RECEIPT_EMAIL'] || 'false'
```
joeyAghion, line 56: This file is a good place to convert configuration to the necessary type (`true`/`false` in this case, rather than a string). You can borrow from the example above that tests for `== 'true'` so that the code locations that reference this can safely expect a boolean.

AlicanAkyuz: thank you, this will make it more straightforward to reference it in the code.

fladson (approved):

config/initializers/_config.rb
```diff
@@ -52,5 +52,6 @@ module Convection
     vibrations_url:
       ENV['VIBRATIONS_URL'] || 'https://admin-partners-staging.artsy.net',
     unleash_url: ENV['UNLEASH_URL'],
-    unleash_token: ENV['UNLEASH_TOKEN']
+    unleash_token: ENV['UNLEASH_TOKEN'],
+    send_new_receipt_email: ENV['SEND_NEW_RECEIPT_EMAIL'] == 'true'
```
patrinoua, line 56: === ? 🤔

kiryl-zubarau (approved):

config/initializers/_config.rb
AlicanAkyuz: `==` suffice in Ruby. [here](https://stackoverflow.com/questions/7156955/whats-the-difference-between-equal-eql-and) is some explanation if you are curious.

PR #1274, opened 2022-03-22
[feat]: allow submission update when user email matches

AlicanAkyuz:
### Description 

Ticket: [[SWA-310](https://artsyproduct.atlassian.net/browse/SWA-310)]

This PR allows updating of a submission if the email address in the request matches the one on the submission.

### DIFF
--- a/app/graphql/resolvers/base_resolver.rb
+++ b/app/graphql/resolvers/base_resolver.rb
@@ -42,6 +42,10 @@ class BaseResolver
       submitted_by_current_session?(submission, session_id)
   end
 
+  def matching_email(submission, user_email)
+    submission.user_email == user_email
+  end
+
   def submitted_by_current_user?(submission)
     submission.user&.gravity_user_id.present? &&
       submission.user&.gravity_user_id == @context&.[](:current_user)
--- a/app/graphql/resolvers/submission_resolver.rb
+++ b/app/graphql/resolvers/submission_resolver.rb
@@ -6,7 +6,9 @@ class SubmissionResolver < BaseResolver
   def run
     check_submission_presence!
 
-    unless draft_in_progress?(submission, @arguments) || admin? || partner?
+    unless draft_in_progress?(submission, @arguments) ||
+             matching_email(submission, arguments&.[](:user_email)) || admin? ||
+             partner?
       raise GraphQL::ExecutionError, 'Submission Not Found'
     end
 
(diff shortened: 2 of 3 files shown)


app/graphql/resolvers/base_resolver.rb
```diff
@@ -42,6 +42,10 @@ def matching_user(submission, session_id)
       submitted_by_current_session?(submission, session_id)
   end
 
+  def matching_email(submission, user_email)
```
fladson, line 45: Usually, methods that return `true` or `false` have a `?` at the end, so I suggest doing it here.
`def matching_email?`

app/graphql/resolvers/submission_resolver.rb
```diff
@@ -6,7 +6,9 @@ class SubmissionResolver < BaseResolver
   def run
     check_submission_presence!
 
-    unless draft_in_progress?(submission, @arguments) || admin? || partner?
+    unless draft_in_progress?(submission, @arguments) ||
+             matching_email(submission, arguments&.[](:user_email)) || admin? ||
```
fladson, line 10: Is `arguments&.[](:user_email)` working as expected?
I would change it to `@arguments&.[](:user_email)` since we are using it as `@arguments` in the first condition `draft_in_progress?(submission, @arguments)`.

app/graphql/resolvers/update_submission_resolver.rb
```diff
@@ -9,7 +9,7 @@ def run
     unless (
              matching_user(submission, @arguments&.[](:session_id)) &&
                submission.draft?
-           ) || admin?
+           ) || matching_email(submission, arguments&.[](:user_email)) || admin?
```
fladson, line 12: [Same comment here.](https://github.com/artsy/convection/pull/1274/files#r833054176)

fladson (commented):
LGTM
Some minor comments and I can help adding some specs if you want.

iskounen (approved):
I agree with @fladson's comments, but otherwise this looks good to me.

app/graphql/resolvers/base_resolver.rb
```diff
@@ -42,6 +42,10 @@ def matching_user(submission, session_id)
...
+    submission.user_email == user_email
```
iskounen, line 46: suggestion: It may be good to `downcase` the email addresses to make sure they match.

PR #1276, opened 2022-03-23
fix(SWA-315) fix published artworks

kiryl-zubarau:
This PR fixes problem with published submissions on My Collection. 

JIRA - https://artsyproduct.atlassian.net/browse/SWA-315

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1288,6 +1288,9 @@ type RemoveAssetFromConsignmentSubmissionPayload {
   clientMutationId: String
 }
 
+"""
+Enum with all submission states for Create or Update Submission
+"""
 enum State {
   APPROVED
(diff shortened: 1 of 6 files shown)


iskounen (commented):
Thanks for highlighting the difference between states that can be used for create/update mutations and all states.

It feels weird to have two enums for the same thing, however. I wonder if we should be using the resolvers to determine whether a create/update mutation is allowed (based on a status) as opposed to having two enums in the schema.

kiryl-zubarau:
@iskounen Agree with you, do you have any thought about what states we should reject on create/update? I started thinking that the enum missing some fields just because the enum wasn't updated in time since there was no needs.

fladson (approved):

PR #1278, opened 2022-03-28
[WIP] Launch My Collection Integration for All Users

iskounen:
This PR removes the check for the `swa_my_collection` Unleash feature flag that occurs before Convection asks Gravity to create an artwork in My Collection when a submission is received. This will effectively launch the My Collection integration for all Artsy users.

The feature flag will still remain in Unleash and in the Convection code after this change, and can be removed at a later date.

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -93,8 +93,7 @@ class SubmissionService
             reject_non_target_supply_artist(submission.artist_id)
           )
           if !submission.draft? && submission.user && !access_token.nil? &&
-               !submission.my_collection_artwork_id &&
-               submission.user&.save_submission_to_my_collection?
+               !submission.my_collection_artwork_id
             create_my_collection_artwork(submission, access_token)
           end
         end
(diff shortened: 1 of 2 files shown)


AlicanAkyuz (approved):

app/services/submission_service.rb
```diff
@@ -93,8 +93,7 @@ def update_submission(
             reject_non_target_supply_artist(submission.artist_id)
           )
           if !submission.draft? && submission.user && !access_token.nil? &&
-               !submission.my_collection_artwork_id &&
-               submission.user&.save_submission_to_my_collection?
```
MounirDhahri, line 97: **thought:** How about we keep this check and instead set the flag value to true for everyone - once we release and everything is fine, we can remove the flag. It's also fine if we remove this as you did since we can redeploy convection as well if something goes wrong

olerichter00: I think removing it should be fine. We can always revert this PR.

olerichter00 (approved):

app/services/submission_service.rb
MounirDhahri: Sure, that sounds good

MounirDhahri (approved):

iskounen:
Thank you all for the review. I am closing this PR for now because I would rather keep the feature flag and change its value/strategy in Unleash to launch this feature.

PR #1279, opened 2022-03-28
[feat]: AlicanAkyuz/swa-310/AddUserToSubmissionMutation

AlicanAkyuz:
### Description 

Ticket: [[SWA-310](https://artsyproduct.atlassian.net/browse/SWA-310)]

This PR creates a new mutation called `AddUserToSubmissionMutation`, which: 
- accepts `submissionId` and `userEmail` as input 
- finds the submission by Id 
- verifies that submission has no `user_id`
- checks `user_email` on submission is matching to `userEmail` from the arguments 
- creates a user and assigns them to the submission 
- creates an artwork in user's myCollection  
- updates the schema with the new mutation


app/graphql/resolvers/add_user_to_submission_resolver.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    #  make sure submission has no user
+    if submission.user_id
+      raise(GraphQL::ExecutionError, 'Submission already has a user')
+    end
+
+    SubmissionService.add_user_to_submission(
+      submission,
+      @context[:current_user],
```
iskounen, line 21: **suggestion:** This assumes that the user requesting this mutation is the owner of the email address that was provided as an argument. While that will be the case for the web and mobile submission flows, I think that we can remove this assumption by providing a `userID` field in the input type. What do you think?

app/services/submission_service.rb
```diff
@@ -311,5 +311,20 @@ def update_submission_info(artwork, submission)
       )
       submission.save!
     end
+
+    def add_user_to_submission(submission, gravity_user_id, access_token)
+      if gravity_user_id.present?
+        user = User.find_or_create_by(gravity_user_id: gravity_user_id)
```
iskounen, line 317: **thought (non-blocking):** It's so confusing to have two sets of user records to manage. I wish this were simpler...

app/services/submission_service.rb
```diff
@@ -311,5 +311,20 @@ def update_submission_info(artwork, submission)
...
+        submission.update!(user_id: user&.id)
+
+        # "send_new_receipt_email" is here for us to test the feature on staging
```
iskounen, line 320: **question:** I don't understand how this helps us test on Staging. Could you provide some more details, please?

iskounen: Oh! That's the name of the config that we're using like a feature flag. I'm sorry for making you explain that 🙏

app/graphql/resolvers/add_user_to_submission_resolver.rb
iskounen: I see what you're saying. A malicious actor could provide a correct `userEmail` that passes validation, while also providing an incorrect `userID` to associate that artwork with a different account.

Since your code works for the currently supported submission flows, let's ignore my suggestion.

iskounen (approved):

PR #1288, opened 2022-04-05
update: there seems to be a problem with the dependabot config here

kajatiger:
In the depandabot config we are referencing a variable for a password that doesn't seem to exist. I am also not sure why we need all these versions listed when we are using bundler.

Bildschirmfoto 2022-04-05 um 18 20 39


.github/dependabot.yml
```diff
@@ -1,87 +1,12 @@
...
-    - 2.0.1
-    - 2.0.2
-  registries:
-  - git-github-com
+  - package-ecosystem: bundler
+    directory: "/"
+    schedule:
+      interval: daily
+    # Limit to 0 to enable only security updates:
+    open-pull-requests-limit: 0
```
kajatiger, line 8: We could also turn it up to something higher.

.github/dependabot.yml
```diff
@@ -1,87 +1,12 @@
...
+    assignees:
+      - jonallured
+    reviewers:
+      - artsy/px-devs
```
kajatiger, line 12: This should be "sell-with-artsy-devs" but I could not find them as a team on github.

joeyAghion: You could create the team if it doesn't already exist? It might be useful beyond these dependabot configs.

joeyAghion (commented):
I don't see private dependencies in this repo's `Gemfile`, so I'm not sure why its Dependabot config ever needed a secret.

.github/dependabot.yml
kajatiger: The parent team "engineering" needs to be approved I think. But I added all team members that I could find the github handle of.

joeyAghion (approved):

PR #1300, opened 2022-04-13
[feat]: email templates with myCollection section

AlicanAkyuz:
### Description

Ticket: [[SWA-338](https://artsyproduct.atlassian.net/browse/SWA-338)]

This PR:
- updates 3 email templates in order to direct users to login/signup on correct artsy web environment  
- updates the paddings in the email templates

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -3,9 +3,14 @@
 class UserMailer < ApplicationMailer
   helper :url, :submissions, :offers
 
+  include UrlHelper
+
   def submission_receipt(submission:, artist:)
     @submission = submission
     @artist = artist
+    @sign_up_url =
+      artsy_formatted_url('signup', { submissionId: @submission.id })
+    @log_in_url = artsy_formatted_url('login', { submissionId: @submission.id })
 
     @utm_params =
       utm_params(
@@ -109,6 +114,10 @@ class UserMailer < ApplicationMailer
   def nsv_bsv_submission_rejected(submission:, artist:)
     @submission = submission
     @artist = artist
+    @sign_up_url =
+      artsy_formatted_url('signup', { submissionId: @submission.id })
+    @log_in_url = artsy_formatted_url('login', { submissionId: @submission.id })
+
     @utm_params =
       utm_params(
         source: 'consignment-rejected',
@@ -125,6 +134,10 @@ class UserMailer < ApplicationMailer
   def non_target_supply_artist_rejected(submission:, artist:)
     @submission = submission
     @artist = artist
+    @sign_up_url =
+      artsy_formatted_url('signup', { submissionId: @submission.id })
+    @log_in_url = artsy_formatted_url('login', { submissionId: @submission.id })
+
     @utm_params =
       utm_params(
         source: 'consignment-rejected',
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -7,6 +7,7 @@ class UserMailer < ApplicationMailer
     @submission = submission
     @artist = artist
     @inject_create_account_section = Convection.config.send_new_receipt_email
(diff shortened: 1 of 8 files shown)


app/views/shared/email/_myCollection_block.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if Rails.env.production? %>
+    <p>Simply <%= link_to 'sign up', "https://www.artsy.net/signup?submissionId=#{@submission.id}", target: :_blank %> or <%= link_to 'log in', "https://www.artsy.net/login?submissionId=#{@submission.id}", target: :_blank %> to Artsy with the same email you used for this submission and visit our app.</p>
+<% else %>
+    <p>Simply <%= link_to 'sign up', "https://staging.artsy.net/signup?submissionId=#{@submission.id}", target: :_blank %> or <%= link_to 'log in', "https://staging.artsy.net/login?submissionId=#{@submission.id}", target: :_blank %> to Artsy with the same email you used for this submission and visit our app.</p>
```
fladson, line 4: The indentation here should be 2, right?

app/views/shared/email/_myCollection_block.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if Rails.env.production? %>
+    <p>Simply <%= link_to 'sign up', "https://www.artsy.net/signup?submissionId=#{@submission.id}", target: :_blank %> or <%= link_to 'log in', "https://www.artsy.net/login?submissionId=#{@submission.id}", target: :_blank %> to Artsy with the same email you used for this submission and visit our app.</p>
```
fladson, line 2: We could also use a helper for this to return just the URL, to avoid duplication.
`<p>Simply <%= link_to 'sign up', "#{SomeHelper.environment_url("signup")}#{@submission.id}", target: :_blank %> or <%= link_to 'log in', "#{SomeHelper.environment_url("login")}#{@submission.id}", target: :_blank %> to Artsy with the same email you used for this submission and visit our app.</p>`

app/views/shared/email/_myCollection_block.html.erb
AlicanAkyuz: right! I'm not sure it affects the rendered html though. I have tried different indentations, but the email template seems to be unaffected by that. 

I'll change it to 2 regardless 👍

app/mailers/user_mailer.rb
```diff
@@ -3,10 +3,14 @@
...
 
+  include UrlHelper
+
   def submission_receipt(submission:, artist:)
     @submission = submission
     @artist = artist
-    @inject_create_account_section = Convection.config.send_new_receipt_email
+    @sign_up_url =
+      artsy_formatted_url('signup', { submissionId: @submission.id })
+    @log_in_url = artsy_formatted_url('login', { submissionId: @submission.id })
```
iskounen, line 13: ❤️

app/mailers/user_mailer.rb
```diff
@@ -3,10 +3,14 @@
 class UserMailer < ApplicationMailer
   helper :url, :submissions, :offers
 
+  include UrlHelper
```
iskounen, line 6: **thought (non-blocking):** I'm surprised to see that you needed to explicitly include `UrlHelper` because I thought that either (a) helpers were globally accessible, or (b) this helper was included in line 4. Mailers and their templates are analogous to Controllers and their views, so I feel like the "correct" thing to do is call the helper method from the template if it is accessible there.

iskounen (approved):
LGTM 🎉

app/mailers/user_mailer.rb
AlicanAkyuz: thank you for this tip 🙇‍♂️

AlicanAkyuz:
received email:

Screenshot 2022-04-13 at 17 53 12

PR #1305, opened 2022-04-20
feat: check user email from gravity instead of args

AlicanAkyuz:
### Description 

This PR improves on security concerns with regards to adding a user to anonymously submitted submissions by: 
- checking the match between the email address on the submission and user's gravity record

### DIFF
--- a/app/graphql/resolvers/base_resolver.rb
+++ b/app/graphql/resolvers/base_resolver.rb
@@ -42,8 +42,9 @@ class BaseResolver
       submitted_by_current_session?(submission, session_id)
   end
 
-  def matching_email(submission, user_email)
-    submission.user_email.downcase == user_email.downcase
+  def matching_email(submission)
+    user = Gravity.client.user(id: @context[:current_user]).user_detail._get
+    submission.user_email.downcase == user.email
   end
 
   def submitted_by_current_user?(submission)
--- a/app/graphql/resolvers/base_resolver.rb
+++ b/app/graphql/resolvers/base_resolver.rb
@@ -47,7 +47,7 @@ class BaseResolver
       user = Gravity.client.user(id: @context[:current_user]).user_detail._get
       submission.user_email.downcase == user.email
     rescue StandardError
-      Rails.logger.info 'Unable match user email with submission email'
+      Rails.logger.info 'Unable to match user email with submission email'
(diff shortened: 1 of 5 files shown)


app/graphql/resolvers/base_resolver.rb
```diff
@@ -42,8 +42,9 @@ def matching_user(submission, session_id)
       submitted_by_current_session?(submission, session_id)
   end
 
-  def matching_email(submission, user_email)
-    submission.user_email.downcase == user_email.downcase
+  def matching_email(submission)
+    user = Gravity.client.user(id: @context[:current_user]).user_detail._get
```
kiryl-zubarau, line 46: Could you please add try...catch here

AlicanAkyuz: I think line 11 in `add_user_to_submission_resolver` would catch it if anything goes wrong here. but I'll add a more specific exception handler 👍

kiryl-zubarau (approved):

app/graphql/resolvers/base_resolver.rb
```diff
@@ -42,8 +42,14 @@ def matching_user(submission, session_id)
...
   end
 
-  def matching_email(submission, user_email)
-    submission.user_email.downcase == user_email.downcase
+  def matching_email(submission)
+    begin
+      user = Gravity.client.user(id: @context[:current_user]).user_detail._get
+      submission.user_email.downcase == user.email
+    rescue StandardError
+      Rails.logger.info 'Unable to match user email with submission email'
```
iskounen, line 50: ❤️

app/graphql/resolvers/base_resolver.rb
```diff
@@ -42,8 +42,14 @@ def matching_user(submission, session_id)
       submitted_by_current_session?(submission, session_id)
   end
 
-  def matching_email(submission, user_email)
-    submission.user_email.downcase == user_email.downcase
+  def matching_email(submission)
+    begin
+      user = Gravity.client.user(id: @context[:current_user]).user_detail._get
+      submission.user_email.downcase == user.email
```
iskounen, line 48: **suggestion:** @AlicanAkyuz should we also `.downcase` the user.email?

PR #1310, opened 2022-04-22
[SWA-320] Instrument convections sidekiq jobs for datadog

fladson:
Let's see if this is enough to send Sidekiq data to datadog.
From what I saw in other repos, this is enough, but we may need to install and configure `gem install dogstatsd-ruby` 
https://artsyproduct.atlassian.net/browse/SWA-320


joeyAghion (approved):
Looks good to me.

PR #1320, opened 2022-05-03
feat: add more delay to SubmissionService notifications

fladson:
This is an attempt to solve the "Still processing images" errors being raised by Sidekiq. Increasing the delay should be enough for the images to be ready.

https://artsyproduct.atlassian.net/browse/SWA-352


iskounen (approved):
LGTM 🎉

PR #1325, opened 2022-06-10
feat: Restforce gem and config

leamotta:
Starting with https://artsyproduct.atlassian.net/browse/PX-5069 to sync approved submissions into Salesforce.

1st step is adding the restforce gem and ENV authentication variables

PR #1327, opened 2022-07-28
feat(PX-5069): Connect approved submissions with Salesforce

leamotta:
https://artsyproduct.atlassian.net/browse/PX-5069

When a consignment submission is approved, sync the information we have about the Artwork and the Contact to Salesforce automatically so that admins don't have to do this process manually.

We're effectively creating an Artwork
Screenshot 2022-07-28 at 18 08 07
Linking it with an Artist
Screenshot 2022-07-28 at 18 08 21
And linking it (or creating and linking) to a Contact
Screenshot 2022-07-28 at 18 08 38


lib/salesforce_service.rb
```diff
@@ -0,0 +1,84 @@
...
+
+class SalesforceService
+  class << self
+    def add_artwork(submission_id)
+      return unless api_enabled?
+
+      submission = Submission.find(submission_id)
+      
+      sf_contact_id = find_contact_id(submission.user_id) || api.create!('Contact', map_submission_to_salesforce_contact(submission))
+      sf_artist_id = api.select('Artist__c', submission.artist_id, ['Id'], 'Gravity_Artist_ID__c').Id
```
leamotta, line 11: can we assume the artist will always be found?

lib/salesforce_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      Convection.config.salesforce_client_id &&
+        Convection.config.salesforce_client_secret &&
+        Convection.config.salesforce_host &&
+        Convection.config.salesforce_username && 
+        Convection.config.salesforce_password &&
+        Convection.config.salesforce_security_token
+    end
+
+    def api
+      @api ||= Restforce.new \
```
leamotta, line 74: Following up on https://github.com/artsy/convection/pull/1325#discussion_r894702957
I don't think it's necessary for our use case volume to go with a connection pool, happy to explore that solution if there are strong feelings about that

lib/salesforce_service.rb
Mitch-Henson: I would say yes, assume the artist will be found otherwise raise. We are already using the bang methods here so it seems like raising is the way we've chosen to go. Also there isn't really a default artist we can use.

lib/salesforce_service.rb
Mitch-Henson: I'm not sure about the current volume of requests, but curious if we could test this out on staging in some way/shape in order to test whether this could impact us?

Mitch-Henson (approved):
Wrote a comment about a potential staging test plan, otherwise looks good to me!

lib/salesforce_service.rb
leamotta: I agree! For reference, the default behaviour of the `select` is to raise (for example, we're catching that in the `find_contact_id` method) so it should be fine as is

lib/salesforce_service.rb
```diff
@@ -0,0 +1,84 @@
...
+        # Other fields we could sync in the future:
+        # Artwork_Status__c: submission.state,
+        # Materials: ???
+        # Diameter: ???
+        # Framed: ???
+        # FramedDimensions: ???
+      }
+    end
+
+    def api_enabled?
```
fladson, line 64: This would pass if we have all fields as empty strings, right? I would put `.present?` in there, just in case.

Another approach would be to raise at boot time.
```
# config/initializers/salesforce.rb
raise "Salesforce credentials missing" unless Convection.config.salesforce_client_id.present?

leamotta: Yes, that's true. Didn't consider it an issue but can add it.

We wouldn't like to raise since for the beginning of the integration this would be deployed and tested a lot on staging, while the production environment remains with empty keys and a "disabled" api state

PR #1331, opened 2022-08-04
fix: Send user gravity id to Salesforce

leamotta:
In Salesforce, the `Contacts` table links to our `Users` through the Gravity user id. This PR changes our SalesforceService to start using the Gravity user id instead of the Convection user id, which is not used externally.

### DIFF
--- a/lib/salesforce_service.rb
+++ b/lib/salesforce_service.rb
@@ -7,7 +7,7 @@ class SalesforceService
 
       submission = Submission.find(submission_id)
       
-      sf_contact_id = find_contact_id(submission.user_id) || api.create!('Contact', map_submission_to_salesforce_contact(submission))
+      sf_contact_id = find_contact_id(submission) || api.create!('Contact', map_submission_to_salesforce_contact(submission))
       sf_artist_id = api.select('Artist__c', submission.artist_id, ['Id'], 'Gravity_Artist_ID__c').Id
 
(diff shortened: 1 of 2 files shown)


lib/salesforce_service.rb
```diff
@@ -7,16 +7,17 @@ def add_artwork(submission_id)
...
       api.create!('Artwork__c', map_submission_to_salesforce_artwork(submission, sf_contact_id, sf_artist_id))
     end
 
     private
     
-    def find_contact_id(user_id)
-      api.select('Contact', user_id, ['Id'], 'Partner_Contact_Ext_Id__c').Id
+    def find_contact_id(submission)
+      return if submission.user.blank?
+      api.select('Contact', submission.user.gravity_user_id, ['Id'], 'Partner_Contact_Ext_Id__c').Id
```
Mitch-Henson, line 20: We are checking for the `user` but is it also possible for the `user.gravity_user_id` to be `nil`?

Mitch-Henson (approved):
Looks good to me! I commented on a potential small change

lib/salesforce_service.rb
leamotta: I think all users should have a gravity user id, however I tested sending `nil` to Salesforce and it works the same (returns a not found error that we are already catching).
Also when we send `nil` at Contact creation time it works fine too!

PR #1334, opened 2022-08-23
feat(PX-5198): Salesforce sync improvements

leamotta:
https://artsyproduct.atlassian.net/browse/PX-5198

- Link primary image and submission id
- Email lookup just in case there is a contact with the same email already in SF


Mitch-Henson (approved):
🚢

PR #1336, opened 2022-08-23
fix(PX-5198): Fix user lookup by email

leamotta:
https://artsyproduct.atlassian.net/browse/PX-5198
If there is no user created and associated with the submission, we can still do a lookup by the user_email to avoid duplicates in SF


Mitch-Henson (approved):
🚢

PR #1343, opened 2022-10-06
fix: Calls SF after saving approval

leamotta:
We're not receiving the proper owner in SF when approving for convection, changing the order of the service calls should solve the issue

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -178,9 +178,9 @@ class SubmissionService
     end
 
     def approve!(submission, current_user)
-      SalesforceService.delay.add_artwork(submission.id)
-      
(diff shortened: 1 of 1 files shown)


app/services/submission_service.rb
```diff
@@ -178,9 +178,9 @@ def submit!(submission)
     end
 
     def approve!(submission, current_user)
-      SalesforceService.delay.add_artwork(submission.id)
-      
       submission.update!(approved_by: current_user, approved_at: Time.now.utc)
+
+      SalesforceService.delay.add_artwork(submission.id)
```
gkartalis, line 183: 👌

gkartalis (approved):

PR #1349, opened 2022-10-13
fix: Authorization workaround for users that exist multiple times in the User table

olerichter00:
Addresses [CX-3058]
## Description

This is an authorization workaround for users that exist multiple times in the User table.

If a user exists twice in Convection’s User table it can break the authentication in the submissions resolver. This PR is just a workaround. We still have to find out why users get created twice and how we can prevent it

[CX-3058]: https://artsyproduct.atlassian.net/browse/CX-3058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/graphql/resolvers/submissions_resolver.rb
+++ b/app/graphql/resolvers/submissions_resolver.rb
@@ -95,9 +95,9 @@ class SubmissionsResolver < BaseResolver
     # TODO: Remove duplicates
     current_users = User.where(gravity_user_id: @context[:current_user])
 
-    !current_users.pluck(:id).include?(
-      base_submissions.select(:user_id).distinct.map { |s| s.user_id }
-    )
+    base_submissions.select(:user_id).distinct.select do |submission|
+      !current_users.pluck(:id).include?(submission.user_id)
+    end.any?
   end
 
   def not_allowed_all_submissions?
--- a/app/graphql/resolvers/submissions_resolver.rb
+++ b/app/graphql/resolvers/submissions_resolver.rb
@@ -91,11 +91,13 @@ class SubmissionsResolver < BaseResolver
     return false if admin? || !@arguments.key?(:ids)
     return true if partner?
 
-    current_user = User.find_by(gravity_user_id: @context[:current_user])
+    # Because there can by more than one user for the currenty user in the DB
+    # TODO: Remove duplicates
+    current_users = User.where(gravity_user_id: @context[:current_user])
 
-    base_submissions.select(:user_id).distinct.map { |s| s.user_id } != [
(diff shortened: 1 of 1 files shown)


app/graphql/resolvers/submissions_resolver.rb
```diff
@@ -91,11 +91,13 @@ def not_allowed_ids?
...
-    current_user = User.find_by(gravity_user_id: @context[:current_user])
+    # Because there can by more than one user for the currenty user in the DB
+    # TODO: Remove duplicates
+    current_users = User.where(gravity_user_id: @context[:current_user])
 
-    base_submissions.select(:user_id).distinct.map { |s| s.user_id } != [
-      current_user&.id
-    ]
+    !current_users.pluck(:id).include?(
+      base_submissions.select(:user_id).distinct.map { |s| s.user_id }
```
lordkiz, line 99: We should be getting the first item in this list instead of passing the entire array, right? `base_submissions.select(:user_id).distinct.map { |s| s.user_id }`

olerichter00: I've updated the logic in the last commit. The first approach didn't work 👍

MounirDhahri (approved):
Sounds good to me, once the PR is merged and staging is ready, please let me know in order to test this out

app/graphql/resolvers/submissions_resolver.rb
```diff
@@ -91,11 +91,13 @@ def not_allowed_ids?
...
-    current_user = User.find_by(gravity_user_id: @context[:current_user])
+    # Because there can by more than one user for the currenty user in the DB
+    # TODO: Remove duplicates
+    current_users = User.where(gravity_user_id: @context[:current_user])
 
-    base_submissions.select(:user_id).distinct.map { |s| s.user_id } != [
-      current_user&.id
-    ]
+    base_submissions.select(:user_id).distinct.select do |submission|
+      !current_users.pluck(:id).include?(submission.user_id)
```
lordkiz, line 99: *Non-blocking*: It may be more performant if we don't have to call `pluck` on `current_users` for each submission.
```suggestion
    current_users_ids = current_users.pluck(:id)
    base_submissions.select(:user_id).distinct.select do |submission|
      !current_users_ids.include?(submission.user_id)
```

olerichter00: Good point! I'll update the code in the case we have to keep this workaround for a longer time.

PR #1351, opened 2022-10-14
Revert "fix: Use image url instead of original image when creating My Collection artwork"

olerichter00:
Reverts artsy/convection#1348


olerichter00:
@MounirDhahri the approach is not working on Staging. I'll revert the change and try to find another solution.

MounirDhahri:
Great, thanks for being on top of this

MounirDhahri (approved):

PR #1356, opened 2022-10-28
Revert "fix: Authorization workaround for users that exist multiple times in the User table"

olerichter00:
The workaround is not needed anymore because we merged all duplicated users.

Reverts artsy/convection#1349


lordkiz (approved):
👍🏾

PR #1360, opened 2022-11-22
fix: change address

dariakoko:
Change floor number from 26 to 24
[CX-3140](https://artsyproduct.atlassian.net/browse/CX-3140)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,5 +1,5 @@
 {
-  "version": "1.3.0",
+  "version": "1.4.0",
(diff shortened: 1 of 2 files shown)


.secrets.baseline
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.3.0",
+  "version": "1.4.0",
```
dariakoko, line 2: that was an automatic change that blocked the commit

MounirDhahri (approved):

PR #1362, opened 2022-11-24
fix(CX-3004): update email address to sell@

MrSltun:
This PR updates the sending email address from consign@ to sell@

### DIFF
--- a/config/initializers/_config.rb
+++ b/config/initializers/_config.rb
@@ -8,7 +8,7 @@ Convection.config =
   OpenStruct.new(
     access_token: ENV['ACCESS_TOKEN'] || 'replace-me',
     admin_names: (ENV['ADMIN_NAMES'] || 'Alice Betty Cindy').split,
-    admin_email_address: ENV['ADMIN_EMAIL_ADDRESS'] || 'consign@artsy.net',
+    admin_email_address: ENV['ADMIN_EMAIL_ADDRESS'] || 'sell@artsy.net',
     artsy_url: ENV['ARTSY_URL'] || 'https://staging.artsy.net',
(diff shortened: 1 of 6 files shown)


MounirDhahri (changes requested):
This looks good to me. There is only one final thing required I believe before merging this.
I see that the email is still consign@artsy in hokusai
Screenshot 2022-11-25 at 10 29 32


It's also still missing within the S3 artsy citadel bucket. Would you mind adding it there

For both instruction, you can follow the steps here  https://artsy.slack.com/archives/CP9P4KR35/p1668617664353739

config/initializers/_config.rb
```diff
@@ -8,7 +8,7 @@ module Convection
   OpenStruct.new(
     access_token: ENV['ACCESS_TOKEN'] || 'replace-me',
     admin_names: (ENV['ADMIN_NAMES'] || 'Alice Betty Cindy').split,
-    admin_email_address: ENV['ADMIN_EMAIL_ADDRESS'] || 'consign@artsy.net',
+    admin_email_address: ENV['ADMIN_EMAIL_ADDRESS'] || 'sell@artsy.net',
```
lordkiz, line 11: Is the env var also updated?
And if so, we also need to confirm it doesn't break stuffs elsewhere.

dariakoko (commented):
LGTM!
Is it all required changes? I suppose we should touch some other repos too, no? 🤔

MrSltun:
@MounirDhahri I just updated the env on both the S3 and on hokusai, please check and let me know :)

MounirDhahri (approved):
Niiice Thanks @MrSltun

PR #1367, opened 2022-12-06
feat(CX-2972): Setup Consignment Inquiries - model, controller, service, route

lordkiz:
This PR resolves [CX-2972]

### Description
In this PR all necessary setup needed for Consignment Inquiries are implemented.
These include:
- Model
- Controller
- Service
- Route
- Tests

#### Follow up
- Add and bind `consignmentinquiry.created` routing key to consignments exchange on rabbitmq 

[CX-2972]: https://artsyproduct.atlassian.net/browse/CX-2972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/api/consignment_inquiries_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    private
+
+    def require_trusted_app_without_current_user
+      has_trusted_app =
+        current_app.present? &&
+          current_user_roles.include?(:trusted)
+      return if has_trusted_app
+
+      raise ApplicationController::NotAuthorized
+    end
```
olerichter00, line 25: **suggestion:** We could use the already existing method from [Api::RestControllser](https://github.com/artsy/convection/blob/5dabc41b0ec00f9e0e1a3c46c2471ae38f0d8095/app/controllers/api/rest_controller.rb#L29).

app/helpers/consignment_inquiries_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ConsignmentInquiriesHelper
+end
```
olerichter00, line 2: **suggestion:** I guess, we could remove this file if it's not needed.

spec/helpers/consignment_inquiries_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
```
olerichter00, line 1: In case you remove the helper class, this spec can be removed as well.

olerichter00 (approved):
Nice work! I've left a few non-blocking suggestions.

app/helpers/consignment_inquiries_helper.rb
lordkiz: Thanks. I will remove it.

app/controllers/api/consignment_inquiries_controller.rb
lordkiz: The existing method requires that `current_user.blank` == true. However for consignment inquiries we want to be able to create whether we have current_user or not.

olerichter00: Hmm, what do you think about renaming the method to `require_trusted_app_and_user` and move it to `Api::RestControllser`? Like this, we can reuse it in the future and it is clear what is requred.

lordkiz: 👍 . But the right name should be `require_trusted_app` since we don't need to reference user in it, but unfortunately that name is already in use and references current_user. 
How about I move this method as it is to Api::RestController?

olerichter00: sounds good to me 👍

app/models/consignment_inquiry.rb
```diff
@@ -0,0 +1,5 @@
+class ConsignmentInquiry < ApplicationRecord
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :name, presence: { message: "is required" }
+  validates :message, presence: { message: "is required" }
```
MounirDhahri, line 4: **thought:** is it possible to check for the trim here?

MounirDhahri: no length check is required, i am referring to avoiding spaces mainly here

lordkiz: Oh. Ok. Yeah, `presence:` does not allow spaces. For eg `message: "-----------space -----------"` will fail validation.

app/controllers/api/rest_controller.rb
```diff
@@ -35,6 +35,15 @@ def require_trusted_app
       raise ApplicationController::NotAuthorized
     end
 
+    def require_trusted_app_without_current_user
+      has_trusted_app =
+        current_app.present? &&
+          current_user_roles.include?(:trusted)
+      return if has_trusted_app
```
joeyAghion, line 42: This doesn't technically do what it says. _Every_ token will have a `current_app` (identified by the `aud` property). If you want to exclude user tokens, you should additionally verify that there's no `current_user` (identified by the `sub` property).

app/models/consignment_inquiry.rb
```diff
@@ -0,0 +1,5 @@
+class ConsignmentInquiry < ApplicationRecord
```
joeyAghion, line 1: Is there a better name for this? "Inquiries" already have a meaning across the platform and I fear this will lead to confusion.

app/controllers/api/rest_controller.rb
lordkiz: I will rename it to `ensure_trusted_app`
This https://github.com/artsy/convection/blob/main/app/controllers/api/rest_controller.rb#L29 here limits choices of what to name the method for when we only want to check trusted_app without bothering about the sub property.

app/models/consignment_inquiry.rb
lordkiz: I actually named it ConsignmentInquiry as a way to distinguish this inquiry from the inquiries we have across platform. I can't think of a better name. 
Any suggestions? 🙏

joeyAghion: I don't have a better suggestion. As long as you tried I guess. 😞

PR #1379, opened 2023-01-05
fix: drop custom slack notifications from CI (unused)

joeyAghion:
Follows up incident #165 and allows SLACK_* CI configs to be removed. Notifications have been suspended since August team/channel shuffles in any case.


sweir27:
👍  Looks good to me once the other issues are fixed.

PR #1386, opened 2023-03-02
Feat(CX-3410): Add Recipient Email to Consignments Inquiries

lordkiz:
Co-authored-by: Daria Koslova <daria.Kozlova@artsymail.com>

This PR partly resolves [CX-3410]

### Description
- We have added an optional recipient_email column to consignment_inquiries. This allows clients to specify the particular collector services team employee that they want to contact in particular. 
- The field also accepts nil. In the event of nil value, we fallback to "sell@artsy.net" on Pulse, thereby not breaking the existing Get in touch flow.

[CX-3410]: https://artsyproduct.atlassian.net/browse/CX-3410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/events/consignment_inquiry_event.rb
+++ b/app/events/consignment_inquiry_event.rb
@@ -13,6 +13,7 @@ class ConsignmentInquiryEvent < Events::BaseEvent
   def properties
     {
       email: @object.email,
+      recipient_email: @object.recipient_email,
       gravity_user_id: @object.gravity_user_id,
       message: @object.message,
       name: @object.name,
--- a/app/events/consignment_inquiry_event.rb
+++ b/app/events/consignment_inquiry_event.rb
@@ -12,8 +12,8 @@ class ConsignmentInquiryEvent < Events::BaseEvent
 
   def properties
     {
-      email: @object.email,
-      recipient_email: @object.recipient_email,
+      email: @object.email, # the email of user sending the inquiry
+      recipient_email: @object.recipient_email, # an optional email of a collector services team member to deliver the inquiry. On Pulse if this is absent, the inquiry will be delivered to sell@artsy.net
       gravity_user_id: @object.gravity_user_id,
       message: @object.message,
       name: @object.name,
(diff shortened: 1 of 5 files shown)


app/events/consignment_inquiry_event.rb
```diff
@@ -13,6 +13,7 @@ def subject
   def properties
     {
       email: @object.email,
+      recipient_email: @object.recipient_email,
```
olerichter00, line 16: **suggestion:** Could we add a comment to explain how `recipient_email` is different to `email`?

app/models/consignment_inquiry.rb
```diff
@@ -1,5 +1,6 @@
 class ConsignmentInquiry < ApplicationRecord
   validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :recipient_email, format: { with: URI::MailTo::EMAIL_REGEXP }, allow_nil: true
```
olerichter00, line 3: **question:** What happens if `recipient_email` is `nil` (because we're allowing `allow_nil: true` here)?

olerichter00 (approved):
Great work 🌟 

Just added some questions

app/events/consignment_inquiry_event.rb
lordkiz: 👍 useful. Thanks. I just did.

app/models/consignment_inquiry.rb
lordkiz: On Pulse when recipient_email the inquiry will be delivered to sell@artsy.net

olerichter00: Makes sense :)

app/events/consignment_inquiry_event.rb
olerichter00: Thanks :)

PR #1402, opened 2023-07-06
fix: Allow partner users to access their submissions by ID

olerichter00:
Addresses [ONYX-22]

[Slack Thread](https://artsy.slack.com/archives/C04KVDZU6/p1688570832817819)

## Description

When loading artworks in My Collection MP does the following request to Vortex in order to get the submission status for submissions that are associated with artworks in the user's collection:

```graphql
query {
  submissions(ids: ["123", "456"]) {
    edges {
      node {
        id
      }
    }
  }
}
```

The query fails if the requesting user also has partner access (e.g., a partner also using their account for My Collection). This is because we restrict the query to users who are not partners if the `ids` param is passed ([GitHub](https://github.com/artsy/convection/blob/27f9ce605d4e1535a68290656b3588eaebd73cf0/app/graphql/resolvers/submissions_resolver.rb#L92)).

This PR adjusts authorization to give users who are also partners access to their submissions by removing the line `return true if partner?`. Removing this line means that partners can use the query, passing the `ids` parameter, to access **only** submissions associated with their (user) account. This is ensured by the following check ([GitHub](https://github.com/artsy/convection/blob/27f9ce605d4e1535a68290656b3588eaebd73cf0/app/graphql/resolvers/submissions_resolver.rb#L98)):

```ruby
user_submissions.map { |s| s.user_id } != [
  current_user&.id
]
```

**Follow Up:**

In addition to this fix, we might want to adjust the MP resolver to recover from a possible failure when querying Convection to fetch the submission metadata.
([GitHub](https://github.com/artsy/metaphysics/blob/54c326e3991f6533623a8001b1332eb7b695f253/src/schema/v2/me/myCollection.ts#L117)).

[ONYX-22]: https://artsyproduct.atlassian.net/browse/ONYX-22?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/requests/api/graphql/queries/submissions_spec.rb
+++ b/spec/requests/api/graphql/queries/submissions_spec.rb
@@ -87,6 +87,28 @@ describe 'submissions query' do
         expect(error_message).to eq "Can't load other people's submissions."
       end
     end
+
+    context 'with a request from a partner and wrong submission id' do
+      let(:token) do
+        JWT.encode(
+          { aud: 'gravity', sub: 'partner_id', roles: 'partner' },
+          Convection.config.jwt_secret
+        )
+      end
+
+      it 'returns an error for that request' do
+        post '/api/graphql', params: { query: query }, headers: headers
+
+        expect(response.status).to eq 200
+        body = JSON.parse(response.body)
+
+        submissions_response = body['data']['submissions']
+        expect(submissions_response).to eq nil
+
+        error_message = body['errors'][0]['message']
+        expect(error_message).to eq "Can't load other people's submissions."
+      end
+    end
   end
 
   describe 'valid requests' do
--- a/app/graphql/resolvers/submissions_resolver.rb
+++ b/app/graphql/resolvers/submissions_resolver.rb
@@ -89,7 +89,6 @@ class SubmissionsResolver < BaseResolver
 
   def not_allowed_ids?
     return false if admin? || !@arguments.key?(:ids)
-    return true if partner?
 
     current_user = User.find_by(gravity_user_id: @context[:current_user])
     user_submissions = base_submissions.select(:user_id).distinct


spec/requests/api/graphql/queries/submissions_spec.rb
```diff
@@ -87,6 +87,28 @@
         expect(error_message).to eq "Can't load other people's submissions."
       end
     end
+
+    context 'with a request from a partner and wrong submission id' do
```
olerichter00, line 91: This test ensures that a partner who does not own the submission (`submission.user_id != partner.user_id`) cannot access the permission by querying for it by ID.

MounirDhahri (approved):
This makes sense to me. thanks for addressing this.
One small comment, you mentioned the request is made to `vortex` in the description, i think you meant convection

MounirDhahri:
Idea for the future:
Since all that we're looking for is if an artwork has been submitted or not, I think it should be fine to probably have that status in Gravity given it's currently a one-time thing and it will save us an additional request to convection looking for all submissions at every request. 

So far, convection has been standing fine, under load but given it's an internal tool, I am not sure it will still be able to handle all the load in the future

dblandin (approved):
LGTM — nice to have that extra spec!

olerichter00:
@MounirDhahri That could be something worth exploring. The idea is to add the submission status to Gravity (associated with the My Collection artwork) and keep it in sync with Convection by updating it whenever the status in Convection changes, right?

The only problem I see is that keeping the status synced between both services could be sensitive to failures.

MounirDhahri:
I would actually suggest not keeping the status in Gravity for the same reasons you mentioned. A flag is all I suggest here. We probably can infer that already now from the existence/non-existence of a submission id for artworks. 

Reason behind that is that we are not using the status anywhere in `MyCollectionQuery` [[Pull Request](https://github.com/artsy/eigen/pull/6210)]

PR #1411, opened 2023-09-08
chore: update renovate config

mc-jones:
This PR updates the assignees of the `renovate.json` to only include current members of Artsy.


amonkhouse (approved):
you are more proactive than me it seems

PR #1428, opened 2024-02-20
chore: Switch to asdf

jonallured:
This PR updates Convection to use ASDF. I attempted to fix all the fallout including things like the Dockerfile but I might have missed something - lmk!

Also I don't know if that node version of 12 (!) moving to 18 is going to be legit but I thought I'd try! Works locally so crossing my fingers.

/cc @artsy/amber-devs


jonallured:
Alright this is green - going to merge and monitor in case I missed something in terms of deployment.

PR #1457, opened 2024-05-20
feat: make Salesforce artwork data available to listing step

joeyAghion:
This is an ugly, incremental step toward making Salesforce's artwork data available to the artwork-listing utility. Most of the effort went into identifying all of the relevant Salesforce fields and mapping them to the corresponding Gravity fields. Still, there are a few fields that need follow-up:
* `Medium_Type__c` is defined in the production Salesforce instance, but not the sandbox. Once added, it's a potential source for the `category` field.
* `Size_of_edition__c` and `Available_works__c` are also not defined in the sandbox, causing submission approvals to fail [when creating the Salesforce artwork](https://github.com/artsy/convection/blob/6518a6d5da268d34a9ca327170e1863e1239f576/lib/salesforce_service.rb#L73-L74).
* The `Edition_Number__c` field is formatted like `13/50` (sometimes), which doesn't directly correspond to Gravity fields.

As ugly as this is, I tried to make this mergeable. The source fields are revealed (for informational purposes) only when clicking a toggle in the listing modal:
list_consignment5

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -281,4 +281,8 @@ class Submission < ApplicationRecord
       metric: dimensions_metric
     }
   end
+
+  def salesforce_artwork
+    @salesforce_artwork ||= SalesforceService.salesforce_artwork_for_submission(self)
+  end
 end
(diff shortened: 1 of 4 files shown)


nickskalkin (approved):
Looks good to me, I would merge like that @joeyAghion!

PR #1459, opened 2024-05-21
feat: add edition set and image source data to listing modal + allow field-by-field choices

joeyAghion:
A little progress pulling edition- and image-related fields into the listing modal.

There are no affordances for picking and choosing different sources by field, but that might come next.

### DIFF
--- a/app/assets/stylesheets/locals.css.scss
+++ b/app/assets/stylesheets/locals.css.scss
@@ -358,3 +358,23 @@ a.make-primary-link {
     height: 500px !important;
   }
 }
+
+#artwork_source_data {
+  tr:nth-child(odd) {
+    background-color: #f9f9f9;
+  }
+
+  label {
+    display: inline-block;
+
+    span,
+    img {
+      padding: 5px;
+    }
+
+    span:has(+input:checked),
+    img:has(+input:checked) {
+      background-color: #eef;
+    }
+  }
+}
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -22,6 +22,8 @@ module Admin
       matching_submissions.page(page).per(size)
     end
 
+    before_action :build_listing_fields, only: %i[show list_artwork]
+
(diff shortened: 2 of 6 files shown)


joeyAghion:
(This whole thing should have `<bad-solution>...</bad-solution>` around it.)

BUT the latest commit allows selecting from either submission or salesforce data on a field-by-field basis. The basic structure is that an `artwork_sources` hash of values like `{"artists": "salesforce", "title": "submission", ...}` is submitted, determining the source of each field. Similarly for `edition_set_sources`. Images are handled a little differently because we want to be able to select any or all of the available images across both sources.

I tested this with a few records but staging data isn't realistic so I'm sure bugs or data incompatibilities will surface later.

joeyAghion:
Looks like:

list_consignment6

joeyAghion:
Rebased and incorporated the recently added Salesforce fields: `Medium_Type__c`, `Size_of_edition__c`, `Available_works__c`

nickskalkin (approved):
Sorry for taking so long to review! I think this PR adds value, sales reps can now choose between different data sources. I think at some point we want to switch convection data that lives in submission model to MyC artwork data (queried from metaphysics).

I would maybe change the markup a bit to be more user friendly. This is a draft for a thing I was building to demo to Devon. Maybe something like this would look better? But up to you, you can leave it like this.

image

joeyAghion:
Oh yes the markup definitely needs improvement, but I didn't feel like I could efficiently accomplish that. 😅 

Do you have that on a branch I could refer to? Or do you want to pair on or push it?

nickskalkin:
@joeyAghion sure, [here](https://github.com/artsy/convection/commit/3871f5364ef61db4a5352ff5ce44d1c5b77b37f2#diff-cee85777dc2a8f34e701330c71eaecb948596af8c5972379cdd430e50a129619R1) is the draft of what I've been working on - it's not ideal, styles are inline in templates, but I guess easily extractable to css. I am happy to pair with you! (on Monday or after preferable)

PR #1460, opened 2024-05-23
feat(onyx-980): setup RabbitMQ consumer

nickskalkin:
In this PR I only setup the RabbitMQ consumer for Gravity's my collection artwork updated event. I use `sneakers` gem like other projects do.

To run sneakers in dev:

```
WORKERS=MyCollectionArtworksWorker bundle exec rails sneakers:run
```

Drafted email (you can try it [on your local](http://localhost:5000/rails/mailers/admin_mailer/artwork_updated)):

image

Wondering if submission fields should be there as well for convenience 🤔  

# Pre-deploy steps

## Staging

- [x] Set missing ENV variables
- [x] Create the queue / exchange and bind them in RabbitMQ management UI

## Production

- [ ] Set missing ENV variables
- [ ] Create the queue / exchange and bind them in RabbitMQ management UI


hokusai/production.yml
```diff
@@ -100,6 +100,91 @@ spec:
                 operator: In
                 values:
                 - foreground
+
```
nickskalkin, line 103: Is there a way to validate these changes before actual deploy? I'm not comfortable doing them blindly.

config/storage.yml
```diff
@@ -1 +1,7 @@
-# Empty file to satisfy ActiveStorage
+local:
```
nickskalkin, line 1: Getting this error when running sneakers, when there is no valid configuration in this file:

```
rails aborted!
KeyError: Missing configuration for the local Active Storage service. Configurations available for the  services.
```

app/mailers/admin_mailer.rb
```diff
@@ -24,4 +24,24 @@ def submission(submission:, user:, artist:)
       subject: "Submission ##{@submission.id}"
     ) { |format| format.html { render layout: "mailer_no_footer" } }
   end
+
+  def artwork_updated(submission:, user_data:, artwork_data:, changes:)
+    assigned_admin = AdminUser.find_by(gravity_user_id: submission.assigned_to)
```
nickskalkin, line 29: Should assigned cataloguer be emailed too?

app/mailers/admin_mailer.rb
```diff
@@ -24,4 +24,24 @@ def submission(submission:, user:, artist:)
...
+    @artwork_id = artwork_data[:id]
+    @changes = changes
+
+    smtpapi category: %w[submission],
+            unique_args: {
+              submission_id: submission.id
+            }
+
+    mail(
+      to: 'nikita.skalkin@artsymail.com', # assigned_admin.email,
```
nickskalkin, line 43: going to switch it after QA on staging & letting secondary team know about this change

app/views/admin_mailer/artwork_updated.html.erb
```diff
@@ -0,0 +1,26 @@
+
+<p>User <%= link_to "##{@user_id}", user_management_url(@submission.user.gravity_user_id) %> (<%= @user_email %>) updated their My Collection artwork associated with the submission you are assigned.</p>
```
nickskalkin, line 2: Any suggestions for more informative message?

hokusai/production.yml
dblandin: You can run hokusai commands locally to create the staging resources before merging. Happy to hop on a call to do that with you.

dblandin (approved):
Looks good! Let's plan on testing/iterating on this in staging.

app/helpers/url_helper.rb
```diff
@@ -39,6 +39,10 @@ def artwork_url(artwork_id)
     "#{Convection.config.artsy_url}/artwork/#{artwork_id}"
   end
 
+  def my_collection_artwork_url(artwork_id)
+    "#{Convection.config.artsy_url}/my-collection/artworks/#{artwork_id}/edit"
```
joeyAghion, line 43: I was surprised to see the plural here, given that the artwork views use `.../artwork/*`, but apparently it's correct. I think we should make them consistent, but that's separate.

app/mailers/admin_mailer.rb
joeyAghion: We might want to return early or send to a fallback address if no assignee is found.

app/services/my_collection_artwork_updated_service.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def initialize(payload)
+    @submission = Submission.find(payload[:object][:submission_id])
+    @subject = payload[:subject]
+    @object = payload[:object]
+    @changes = payload[:properties][:changes]
+  end
+
+  def notify_admin!
+    return if ENV["ENABLE_MYC_ARTWORK_UPDATED_EMAIL"] != "true"
```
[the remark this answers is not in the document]
joeyAghion: +1 to consolidating any configuration in the `_config.rb` initializer.

hokusai/production.yml
joeyAghion: Also `hokusai staging update --dry-run --skip-checks ...`

joeyAghion (commented):
Late comments incoming.

And a note about the `WORKERS` environment variable: we previously found that it was confusing to have to list all of the desired Sneakers workers in that variable (see [PLATFORM-3369](https://artsyproduct.atlassian.net/browse/PLATFORM-3369)), so switched to leaving it empty which has the effect of loading _all_ defined workers. We should do that in Convection too.

PR #1467, opened 2024-05-28
feat: notify admin about MyC image uploads

nickskalkin:
I render all types of images to give admin more freedom to choose from, but probably large or normalized image is enough? 

Screenshot of an example email:

image


nickskalkin:
Going to merge because the change is behind FF - going to address code review requests in follow-up PR @dblandin

PR #1469, opened 2024-06-04
fix "could not obtain a connection from the pool" error

nickskalkin:
Should fix "could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use" error for sneakers. Default threads count is 10 as per [doc](https://github.com/jondot/sneakers/wiki/Configuration#workers), but `RAILS_MAX_THREADS` is 5.

Also made a change to notify admins only about changes made to approved submissions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ latest.dump
 node_modules/
 public
 yarn-error.log
+.ruby-version
--- a/app/services/my_collection_artwork_updated_service.rb
+++ b/app/services/my_collection_artwork_updated_service.rb
@@ -16,7 +16,7 @@ class MyCollectionArtworkUpdatedService
       return
     end
 
-    if submission.state != "submitted"
+    if submission.state != "approved"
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ latest.dump
 node_modules/
 public
 yarn-error.log
+.ruby-version
```
nickskalkin, line 14: `.ruby-version` was removed from this repo when switching to asdf. I still use rbenv though and find it useful to have `.ruby-version` in the repo.

dblandin (approved):
Looks like we set the default to `5` as well for the connection pool [here](https://github.com/artsy/convection/blob/819523345cc245b8a797429071be29c0aff05cc6/config/database.yml#L4).

Diff looks good but ideally we'd have two separate commits to merge in:
1) fix: could not obtain a connection from the pool
2) fix: only send admin updated notification for approved submissions

Think we can restructure the commits before merge?

nickskalkin:
makes sense! did it @dblandin

joeyAghion:
🙏

PR #1471, opened 2024-06-05
feat(onyx-1042): submission GraphQL type supports myCollectionArtworkID field

nickskalkin:
To more easily query for artwork metadata stored in My Collection via an associated Submission (in Convection), we want to expose an artwork field under Submission within the GraphQL schema:

```graphql
query X {
  submission(id: "x") {
    artwork {
      ...
    }
  }
}
```

Convection needs to return myCollectionArtworkID field so that Metaphysics could do the stitching thing.

### DIFF
--- a/app/graphql/types/submission_type.rb
+++ b/app/graphql/types/submission_type.rb
@@ -36,6 +36,7 @@ module Types
     nilable_field :location_country_code, String, null: true
     nilable_field :medium, String, null: true
     nilable_field :minimum_price_dollars, Integer, null: true
+    nilable_field :myCollectionArtworkID, String, null: true, method: :my_collection_artwork_id
     nilable_field :primary_image, Types::AssetType, null: true
     nilable_field :provenance, String, null: true
     nilable_field :published_at, GraphQL::Types::ISO8601DateTime, null: true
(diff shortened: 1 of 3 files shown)


app/graphql/types/submission_type.rb
```diff
@@ -36,6 +36,7 @@ class SubmissionType < Types::BaseObject
     nilable_field :location_country_code, String, null: true
     nilable_field :medium, String, null: true
     nilable_field :minimum_price_dollars, Integer, null: true
+    nilable_field :myCollectionArtworkID, String, null: true, method: :my_collection_artwork_id
```
joeyAghion, line 39: To confirm, the capitalization used here (and in `sourceArtworkID` below) is preferred over the style of all the other fields?

nickskalkin: I think `sourceArtworkID` was explicitly written like that (with ID both capital letters), because otherwise `source_artwork_id` would be compiled as `sourceArtworkId` (d is a lowercase letter). I decided to follow this "convention"

joeyAghion (approved):

olerichter00 (approved):
L

PR #1480, opened 2024-06-21
feat: update logic for associating a submission to the logged in user

dblandin:
Before:
- Submission email address must match logged in user's email address
- Submission user_id must not already be present

After:
- If user_id is present, it must corresponding to the ID of the logged in user
- If user_id is not present, submission must be in a draft state

cc/ @joeyAghion as you were interested in these permission changes.

### DIFF
--- a/app/graphql/resolvers/add_user_to_submission_resolver.rb
+++ b/app/graphql/resolvers/add_user_to_submission_resolver.rb
@@ -27,4 +27,8 @@ class AddUserToSubmissionResolver < BaseResolver
 
     {consignment_submission: submission}
   end
+
+  def submission
+    @submission ||= Submission.find_by(uuid: submission_id)
+  end
 end
--- a/app/graphql/resolvers/add_user_to_submission_resolver.rb
+++ b/app/graphql/resolvers/add_user_to_submission_resolver.rb
@@ -6,14 +6,17 @@ class AddUserToSubmissionResolver < BaseResolver
   def run
     check_submission_presence!
 
-    # check if user's email in Gravity matches the one on the submission
-    unless matching_email(submission)
-      raise(GraphQL::ExecutionError, "Submission not found for this user")
+    # early no-op if already claimed by the current user
+    if submitted_by_current_user?(submission)
+      return {consignment_submission: submission}
     end
 
-    #  make sure submission has no user
     if submission.user_id
       raise(GraphQL::ExecutionError, "Submission already has a user")
+    end
+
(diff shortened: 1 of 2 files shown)


dblandin:
@olerichter00 @nickskalkin This should be ready for review. Added some missing test coverage to exercise the new behavior.

app/graphql/resolvers/add_user_to_submission_resolver.rb
```diff
@@ -6,23 +6,25 @@ class AddUserToSubmissionResolver < BaseResolver
...
     end
 
-    #  make sure submission has no user
     if submission.user_id
       raise(GraphQL::ExecutionError, "Submission already has a user")
     end
 
+    if submission.state != Submission::DRAFT
+      raise(GraphQL::ExecutionError, "Submission must be in a draft state to claim")
+    end
```
olerichter00, line 20: 👍

olerichter00 (approved):
The code looks good 👍 

I'm just curious why the email check isn't needed anymore. 🤔

spec/requests/api/graphql/mutations/add_user_to_submission_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        mutation {
+          addUserToSubmission(input: { id: \"#{submission.id}\" }) {
+            clientMutationId
+            consignmentSubmission {
+              internalID
+            }
+          }
+        }
+    GRAPHQL
+
```
nickskalkin, line 57: **suggestion**: I would add something like `expect(SubmissionService).to receive(:add_user_to_submission).never` to confirm that it actually no-ops

nickskalkin (approved):
👍

dblandin:
> I'm just curious why the email check isn't needed anymore.

Since removing the Contact Information step, we're no longer collecting an unauthenticated visitor's email address in the submission flow — so the current email check can no longer be satisfied. At least that's my current understanding.

app/graphql/resolvers/add_user_to_submission_resolver.rb
```diff
@@ -6,23 +6,25 @@ class AddUserToSubmissionResolver < BaseResolver
   def run
     check_submission_presence!
```
joeyAghion, line 7: [Suggested a change here](https://artsy.slack.com/archives/C05EQL4R5N0/p1719515049539169?thread_ts=1718979219.264869&cid=C05EQL4R5N0).

PR #1481, opened 2024-06-25
feat: support total_count param for GET /submissions

MounirDhahri:
This PR adds support to a new param to `GET /submissions` endpoint in order to return the total count of submission.
This might not be the best approach because we might need to support the same param in other endpoints, It would be interesting to investigate how to support that everywhere - but maybe that's not possible.

Why did we add this?
This was important in order to return the `totalCount` of submissions for a user in the `submissionsConnection` interface.

### DIFF
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -37,7 +37,7 @@ module Api
 
       # Return x-total-count header if total_count param is present. Keep this last
       if params[:total_count].present?
-        headers["X-Total-Count"] = Submission.count
+        headers["X-Total-Count"] = submissions.count
       end
 
       render json: submissions.to_json, status: :ok
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -24,6 +24,10 @@ module Api
     end
 
     def index
+      # Return x-total-count header if total_count param is present
+      if params[:total_count].present?
+        headers["X-Total-Count"] = Submission.count
+      end
       param! :completed, :boolean, default: nil
 
       user = User.where(gravity_user_id: current_user).first
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -24,10 +24,6 @@ module Api
     end
 
     def index
-      # Return x-total-count header if total_count param is present
-      if params[:total_count].present?
-        headers["X-Total-Count"] = Submission.count
-      end
       param! :completed, :boolean, default: nil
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/submissions_controller.rb
```diff
@@ -24,6 +24,10 @@ def update
     end
 
     def index
+      # Return x-total-count header if total_count param is present
+      if params[:total_count].present?
+        headers["X-Total-Count"] = Submission.count
```
dblandin, line 29: **suggestion**: This would always return the total number of Submissions in the database. We should push this further down below any scopes/filters, if we need it. What UI would this support?

MounirDhahri: > This would always return the total number of Submissions in the database. We should push this further down below any scopes/filters

Good catch, will update

> What UI would this support?

I added this to keep the API the same as we do with Gravity for connections
Screenshot 2024-06-25 at 14 00 55

app/controllers/api/submissions_controller.rb
```diff
@@ -34,6 +34,12 @@ def index
       end
 
       submissions = submissions.order(created_at: :desc).page(page).per(size)
+
+      # Return x-total-count header if total_count param is present. Keep this last
+      if params[:total_count].present?
+        headers["X-Total-Count"] = Submission.count
```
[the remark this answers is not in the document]
nickskalkin: Oh actually just `submissions.count`, because there are other filters above and it is already filtered by user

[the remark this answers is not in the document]
MounirDhahri: good catch - thanks!

app/controllers/api/submissions_controller.rb
```diff
@@ -34,6 +34,12 @@ def index
       end
 
       submissions = submissions.order(created_at: :desc).page(page).per(size)
+
+      # Return x-total-count header if total_count param is present. Keep this last
```
nickskalkin, line 38: `Keep this last` - was this sentence incomplete?

spec/requests/api/submissions/index_spec.rb
```diff
@@ -57,6 +57,18 @@
       expect(body.first["id"]).to eq submission.id
     end
 
+    fit "returns total count when total_count param is present" do
```
nickskalkin, line 60: don't forget to remove `fit`!

nickskalkin (approved):
looks good, left 2 small comments

MounirDhahri:
Closing this for now - we can bring it again if we decide to support save and continue in the backend

PR #1482, opened 2024-06-25
feat: add state filtering to Get /submissions

MounirDhahri:
This PR adds state filtering to `Get /submissions` endpoint. This comes as a follow-up on 
https://github.com/artsy/metaphysics/pull/5811#pullrequestreview-2138312442

### DIFF
--- a/spec/requests/api/submissions/index_spec.rb
+++ b/spec/requests/api/submissions/index_spec.rb
@@ -31,22 +31,20 @@ describe "Show Submission" do
       expect(body).to eq []
     end
 
-
-    fit "returns only draft submissions when state is draft" do
-      submission = Fabricate(:submission, user: user, state: "draft")
+    it "returns only draft submissions when state is draft" do
+      Fabricate(:submission, user: user, state: "draft")
       Fabricate(:submission, user: user, state: "approved")
       Fabricate(:submission, user: user, state: "submitted")
 
       get "/api/submissions",
         params: {
-          state: "draft"
+          state: ["draft", "submitted"]
         },
         headers: headers
 
       expect(response.status).to eq 200
       body = JSON.parse(response.body)
-      expect(body.length).to eq 1
-      expect(body[0]["id"]).to eq submission.id
+      expect(body.length).to eq 2
     end
 
     it "returns your own submissions" do
--- a/spec/requests/api/submissions/index_spec.rb
+++ b/spec/requests/api/submissions/index_spec.rb
@@ -45,6 +45,7 @@ describe "Show Submission" do
       expect(response.status).to eq 200
       body = JSON.parse(response.body)
       expect(body.length).to eq 2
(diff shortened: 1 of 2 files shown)


spec/requests/api/submissions/index_spec.rb
```diff
@@ -31,6 +31,24 @@
       expect(body).to eq []
     end
 
+
+    fit "returns only draft submissions when state is draft" do
```
dblandin, line 35: **suggestion**: should `s/fit/it` otherwise all of the specs in this file will be skipped.

spec/requests/api/submissions/index_spec.rb
```diff
@@ -31,6 +31,24 @@
...
+      submission = Fabricate(:submission, user: user, state: "draft")
+      Fabricate(:submission, user: user, state: "approved")
+      Fabricate(:submission, user: user, state: "submitted")
+
+      get "/api/submissions",
+        params: {
+          state: "draft"
```
dblandin, line 42: **suggestion**: As implemented, the API endpoint should also support an array of state values to filter by but it would be nice to have a separate spec here to exercise that. I imagine that the GraphQL addition would also support one or more values to filter by.

```ruby
get "/api/submissions",
  params: {
    state: ["submitted", "approved"]
  },
  headers: headers
```

dblandin (commented):
Looking good!

spec/requests/api/submissions/index_spec.rb
MounirDhahri: I think I left a comment about this somewhere else. Would it make sense to have the state as an array here? I see we have both patterns of `state` and `states` in MP.

dblandin: I think `state` is fine. REST clients can pass a single value `state=draft` or multiple values `state[]=submitted&state[]=approved`.

For MP/GraphQL, I also think multiple values can be supported:

```graphql
{
  submissionsConnection(first: 1, state: [DRAFT])
  submissionsConnection(first: 1, state: [SUBMITTED, APPROVED])
}
```

spec/requests/api/submissions/index_spec.rb
```diff
@@ -31,6 +31,22 @@
...
+
+      get "/api/submissions",
+        params: {
+          state: ["draft", "submitted"]
+        },
+        headers: headers
+
+      expect(response.status).to eq 200
+      body = JSON.parse(response.body)
+      expect(body.length).to eq 2
```
olerichter00, line 47: Maybe we can add a test to check if the correct submissions are returned. 

```suggestion
      expect(body.length).to eq 2
      expect(body.map { |s| s.state ).to eq ["draft", "submitted"]
```

olerichter00 (approved):
Looks good 🌟

spec/requests/api/submissions/index_spec.rb
MounirDhahri: good thought!

PR #1483, opened 2024-06-25
fix: Adjust update submission error message when state not draft

olerichter00:
Context: https://github.com/artsy/force/pull/14076#discussion_r1652855229

## Description

Adjust update submission error message when state not draft.


dblandin (approved):
Nice 💯

We can build on this as requirements around updating change but this nicely improves the error messaging for the current state of things 👍🏼

PR #1487, opened 2024-07-08
feat(onyx-1087): new asset type 'additional files'

nickskalkin:
[Jira](https://artsyproduct.atlassian.net/browse/ONYX-1087)

This PR adds new asset type "additional_file", which allows to upload any type of files (not limited to images).

Admins users can upload / download such files via admin interface:

image

image

This PR also adds Metaphysics support for new asset type - exposes new mutation arguments (s3Path, s3Bucket) to allow clients to specify where the file is. S3 location is not accessible to public users, therefore Convection pre-signs asset url and exposes it as a new `documentPath` property.

# Deploy

- [ ] set `AWS_UPLOAD_BUCKET` env variable on staging / prod

# For reviewers

Attaching here a couple of notes to help you to navigate through the code, since this PR includes multiple changes.

## File upload

Main changes related to file upload are in this file: https://github.com/artsy/convection/pull/1487/files#diff-166a22f3aef8002c5064540e25e2c427bd29732202af971ffbb6e58238c6d366R1. For images, Convection uses our own https://github.com/artsy/gemini_upload-rails gem, which is a wrapper around Gemini and https://github.com/blueimp/jQuery-File-Upload. This gem dynamically prepares the form for s3 upload - it fetches upload configuration from Gemini and then submits the form using this jquery-file-uload library.

I do it similar thing for the new asset type, but there is no need to go to Gemini, so the form config is rendered on backend:

- https://github.com/artsy/convection/pull/1487/files#diff-00832b81dd15c96731fd4ebd9edf330ec827b3add08b119506925c6b68a7aefcR1
- https://github.com/artsy/convection/pull/1487/files#diff-bd72bc7edd5884dc9422088bb37aebfab182ba415384e2106ce6701ce473ce0dR1

Frontend passes `additional_file_keys` and `names`, which backend uses to create assets here: https://github.com/artsy/convection/pull/1487/files#diff-52d98c35c625099704543c387ac70e728efb04bfa7125d73bd81109907ae451fR3

## Access to files

Files are then accessible via pre-signed urls (urls expire in some minutes for security reasons):

- https://github.com/artsy/convection/pull/1487/files#diff-0545770a3fbf7606692cd849fe6ca5dd4b454b9a203b8588657abbe7f0ea30b8R57
- https://github.com/artsy/convection/pull/1487/files#diff-412023f14b19c2efa5789da4580ddff970d21d92e54e7e40680b80b64ffac1fbR19

## GraphQL

For "create asset" mutations, I made `gemini_token` optional (add added new s3 location fields):

- https://github.com/artsy/convection/pull/1487/files#diff-1c8066499670f32295ab501950c5ace240dccb482738c1b403b7ac95d7974856R5
- https://github.com/artsy/convection/pull/1487/files#diff-9792b956859f76409c9cf05fd151c6885a83789f032c35f9798525c779324a9cR5

To create an "additional_file" via GraphQL you can execute such mutation:

```graphql
mutation {
  addAssetToConsignmentSubmission(input: { submissionID: 1, filename: "name.pdf", s3Path: "path", s3Bucket: "bucket"}) {
    clientMutationId
    asset {
      id
      submissionId
    }
  }
}

```

Assets are now filterable: https://github.com/artsy/convection/pull/1487/files#diff-124d27f404b8d8bd93de7d900554f4d59e4f749b277797cd315a63d9df024fbfR20

Since clients already do `submission.assets` and expect only images in response - I filter by images by default. It's possible to fetch additional files like this:

```graphql
{
  submission(id: "246195") {
    id
    assets(assetType: [ADDITIONAL_FILE]) {
      id
      filename
      imageUrls
      documentPath
    }
  }
}
```


lib/s3.rb
```diff
@@ -0,0 +1,21 @@
+class S3
```
olerichter00, line 1: https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do 🤷 

```suggestion
# frozen_string_literal: true

class S3
```

olerichter00 (commented):
🌟

app/graphql/mutations/add_assets_to_consignment_submission.rb
```diff
@@ -2,13 +2,15 @@
...
-    argument :gemini_tokens, [String], required: true
+    argument :gemini_tokens, [String], required: false
     argument :submissionID, ID, required: false
     argument :external_submission_id, ID, required: false
     argument :sessionID, String, required: false
     argument :asset_type, String, required: false
     argument :filename, String, required: false
     argument :size, String, required: false
+    argument :s3_buckets, [String], required: false
+    argument :s3_paths, [String], required: false
```
dblandin, line 13: Thoughts on supporting an extra object type for an `UploadSource` to encapsulate the key+bucket pairs?

Thinking of something like:

```graphql
{
  addAssetToConsignmentSubmission(input: {
    source: { key: "KEY", bucket: "BUCKET" }
  }) {
    //
  }

  addAssetsToConsignmentSubmission(input: {
    sources: [{ key: "KEY", bucket: "BUCKET" }]
  }) {
    //
  }
}
```

app/controllers/admin/assets_controller.rb
```diff
@@ -3,7 +3,7 @@
 module Admin
   class AssetsController < ApplicationController
     before_action :set_submission
-    before_action :set_asset, only: %i[show destroy]
+    before_action :set_asset, only: %i[show destroy download]
```
dblandin, line 6: **question**: Does this new `download` action still exist?

app/graphql/mutations/add_assets_to_consignment_submission.rb
dblandin: Also noticed that we tend not to encode "s3" into field names in Metaphysics. Probably in an effort not to lock the schema to any one provider. So might be worth choosing more generic names here as well.

dblandin (commented):
Looking good!

One other thing to think about later is how clients currently upload assets to S3. It looks like clients fetch credentials from Gemini and then use those credentials to upload the source media before creating the "Asset" in our systems.

We might be able to continue "borrowing" that Gemini-backed authorization but would be worth thinking about having a similar field via Convection's API to provide upload instructions/credentials to clients for generic file uploading.

app/assets/javascripts/new_asset.js.coffee
```diff
@@ -1,51 +1,124 @@
...
+  $(".uploading_file[data-uid=#{uid}] progress").remove()
+
+addProgressLine = (e, data) ->
+  uid = data.uid
+  fileName = data.files[0].name
+  uploadPercent = "<div><progress value='0' max='100' class='upload_progress_bar'>0</progress></div>"
+  $('#uploading_files').append("<div class='uploading_file' data-uid=#{uid}><div class='uploading-file-meta clearfix'><div class='uploading-file-name pull-left'>#{fileName } <em><span class='upload_percentage'>0</span>%</em></div></div>#{uploadPercent}</div>")
+
+onFailedUpload = () ->
+  $('#uploading_failed').html('Failed to upload image.')
```
joeyAghion, line 40: Maybe just "asset" or "file" instead.

joeyAghion (approved):
Non-blocking thought: it might be useful to distinguish _specific_ types of additional files, like invoices, certificates of authenticity, or "other." Knowing that those files have been uploaded at a glance (without downloading them) could be useful for administrators, and those file types could all operate exactly like the "additional file" introduced here. Without that, we're relying on the uploaded file name to be descriptive enough.

app/controllers/admin/assets_controller.rb
nickskalkin: no, forgot to remove it, thanks

nickskalkin:
feedback question @olerichter00 @dblandin @joeyAghion: was "code walkthrough" in PR description useful (did it make code review easier) or I just waster mine and yours time?

dblandin:
Definitely not a waste of time! I found the notes useful to orient myself _before_ jumping into the code. Very much a fan of thorough PR descriptions to offer that context.

Although once I jump to the diff mode, I'm unlikely to switch back to the PR description so I think file or line-specific "code tour" comments might have been more useful to have in the diff as comments rather than in the PR description.

olerichter00:
> feedback question @olerichter00 @dblandin @joeyAghion: was "code walkthrough" in PR description useful (did it make code review easier) or I just waster mine and yours time?

Always a good idea! And I agree with Devon, and maybe one entry point and links from comment to comment could make it easier to stay in the diff mode.

PR #1490, opened 2024-07-11
feat(onyx-1109): allow users to edit non-draft submissions

nickskalkin:
Allow users to edit non-draft submissions


olerichter00 (approved):
Thanks! 🙏

PR #1491, opened 2024-07-11
feat: track history of submission changes

joeyAghion:
This is a "spike" into integrating `paper_trail` for lightweight tracking of submission changes (related to [ONYX-1109](https://artsyproduct.atlassian.net/browse/ONYX-1109)). I followed [the default setup instructions](https://github.com/paper-trail-gem/paper_trail), and just overrode the `user_for_paper_trail` method for our 2 different use-cases (administrative requests, or API requests including via GraphQL). No user is tracked for logged-out updates, but that should be obvious from the submission data, I guess.

I don't want to bloat the repo for no reason, but if this sort of history is useful, we could consider a few further updates, neither difficult:
* make the history available in the UI (in some raw form)
* consider [switching](https://github.com/paper-trail-gem/paper_trail?tab=readme-ov-file#6b-custom-serializer) from YAML (the default) to JSON data formatting

[ONYX-1109]: https://artsyproduct.atlassian.net/browse/ONYX-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem "jquery-rails"
 gem "kaminari" # for pagination
 gem "money" # for currency/money formatting
 gem "neat", "1.7.2" # required for watt
+gem "paper_trail" # track history of submissions
 gem "pg_search" # for searching within convection's database
 gem "premailer-rails" # generate text parts from HTML automatically
 gem "rack-cors" # to allow cross-origin requests
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -327,6 +327,9 @@ GEM
     omniauth-rails_csrf_protection (1.0.0)
       actionpack (>= 4.2)
       omniauth (~> 2.0)
+    paper_trail (15.1.0)
+      activerecord (>= 6.1)
+      request_store (~> 1.4)
     parallel (1.20.1)
     parser (3.0.1.1)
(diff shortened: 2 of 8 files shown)


nickskalkin (commented):
The change looks good to me, but here is what I don't fully understand. Once the submission is approved it has a corresponding my collection artwork (maybe this is not true for all submissions, but for all user-generated ones). And we already track artwork changes in gravity. Also, there are some fields in my collection artworks table, that do not exist in submissions (tier 2 information like framed metrics and some others). The question is - maybe it is enough just to track MyC artwork changes and expose it in some way for convection admins? (either Tale or even by using [this recent addition](https://github.com/artsy/gravity/blob/main/app/models/concerns/my_collection_artwork.rb#L112))

At the same time I think that your approach is much simper from technical point of view.

joeyAghion:
Yes, fair point. I thought we were mostly discussing submission changes (like additional file uploads). We probably don't need both, but if we'll allow updates to submissions at all, this could be useful.

joeyAghion:
I'll close this... it sounds like we're on the same page that we can easily track changes, should that de-risk our product changes.

PR #1493, opened 2024-07-16
feat(onyx-1099): add address line 1 & 2 to submissions

nickskalkin:
[Jira](https://artsyproduct.atlassian.net/browse/ONYX-1099)

Adds address line 1 & 2 for tier 2 information needs. 
Adds these fields to graphql object type and to create / update mutations

Sample request 1:

```graphql
{
  submission(id: "246195") {
    id
    locationAddress
    locationAddress2
    locationCity
  }
}
```

Sample request 2:

```graphql
mutation {
  updateConsignmentSubmission(input: { id: "246195", locationAddress: "Not fake str", locationAddress2: "apt 3" }) {
    consignmentSubmission {
      locationAddress
      locationAddress2
    }
  }
}
```

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -534,6 +534,8 @@ input CreateSubmissionMutationInput {
   editionSize: Int
   editionSizeFormatted: String
   height: String
+  locationAddress: String
+  locationAddress2: String
   locationCity: String
   locationCountry: String
   locationCountryCode: String
@@ -1374,6 +1376,8 @@ type Submission {
   """
   id: ID!
   internalID: ID
+  locationAddress: String
+  locationAddress2: String
   locationCity: String
   locationCountry: String
   locationCountryCode: String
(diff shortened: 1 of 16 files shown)


_schema.graphql
```diff
@@ -534,6 +534,8 @@ input CreateSubmissionMutationInput {
   editionSize: Int
   editionSizeFormatted: String
   height: String
+  locationAddress: String
+  locationAddress2: String
```
olerichter00, line 538: What do you think about renaming this to `addressLine` to match the naming of Gravity's [UserAddress type](https://github.com/artsy/metaphysics/blob/10863e6753235b3286754d891db787471defbd24/_schemaV2.graphql#L20590).

```suggestion
  locationAddressLine1: String
  locationAddressLine2: String
```

olerichter00 (approved):
Thanks a lot for the support 🌟

_schema.graphql
nickskalkin: Sure, I'll do it. We were inspired by gravity's `address` and `address_2` but since MP converts it to `locationAddressLine` I think it's better to follow this convention

MounirDhahri: we tried to match how it is in the Location type 
https://github.com/artsy/metaphysics/blob/585fff241d6291c65be31e31a8d6767f18c4a29f/_schemaV2.graphql#L11339-L11342

olerichter00: Actually, both versions would work fine. My Collection stores the artwork location as [MyLocation](https://github.com/artsy/force/blob/726b38534a8ea68e80db4fa0f8e71bf45b7b96e6/data/schema.graphql#L14129), which uses `address` and `address2`.

PR #1497, opened 2024-07-25
feat(onyx-1104): add new submission status 'resubmitted'

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-1104)

Adds a new 'resubmitted' status. Resubmitted is when submission was already approved and then user added tier 2 details (additional documents, frame info, condition, and location). 

What this PR does not cover is sync new data with Salesforce. One of the reasons is that some tier 2 data (frame info, condition) lives in Gravity (artwork model), which complicated sync a bit. I've created a [Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-1136) for this. 

# TODO

- [ ] Sync schema with MP


app/models/submission_state_actions.rb
```diff
@@ -15,6 +15,8 @@ def run
     case submission.state
     when Submission::DRAFT, Submission::SUBMITTED
       [approve_action, publish_action, hold_action, reject_action, close_action]
+    when Submission::RESUBMITTED
+      [publish_action, hold_action, reject_action, close_action]
```
dblandin, line 19: Noticing that we don't allow a submission to transition to `rejected` once it has been `approved`. Since this new `resubmitted` state comes after `approved`, should the same rule apply? Perhaps the only allowed options should be publish/hold/close.

dblandin (approved):
One minor comment but overall LGTM!

app/models/submission_state_actions.rb
nickskalkin: Good question, is it possible that info provided in tier 2 "invalidates" the whole submission? For instance, user initially specified location as "Germany" (in tier 1), but then specified some country in tier 2 from which we cannot organize artwork delivery. I'm just speculating - don't know if such case is real.

Also user could change something even in tier 1 information list - I think in this cases we should allow rejecting such submissions

dblandin: In that case, maybe the submission is `closed` instead? I feel like the approval/rejection decision falls at the same point in the internal process and it might be awkward to have a submission go from submitted > approved > rejected based on how that process is currently defined.

PR #1501, opened 2024-07-29
feat: Update My Collection artwork when creating a submission

olerichter00:
Addresses [ONYX-1133]

## Description

Currently, it is not possible to query draft submissions (when the Tier 1 submission flow is not yet finished) on a My Collection artwork ([GitHub](https://github.com/artsy/metaphysics/blob/3390f03de49bc99765ba6d62cb283a1b2a085753/src/schema/v2/artwork/index.ts#L429)), because they get linked when the submission gets submitted (state change to "SUBMITTED"). For the new My Collection submission status design, this would be necessary to make draft submissions visible on the My Collection artwork page.

---

With this PR, the submission will be linked to the My Collection artwork already when the submission is created. This would ensure that the artwork and the submission are linked correctly when a user submits an artwork from My Collection.

---

**Alternative Approaches**

Alternatively, we could update the My Collection artwork in the [stitched Metaphysics resolver](https://github.com/artsy/metaphysics/blob/3390f03de49bc99765ba6d62cb283a1b2a085753/src/lib/stitching/convection/v2/stitching.ts#L129). My reasoning for doing it in Convection instead is that we would then have two different places where we update the related My Collection artwork (Convection & Metaphysics).  

In the future, it might be good to migrate these updates to Metaphysics and not update data in Gravity from Convection.

[ONYX-1133]: https://artsyproduct.atlassian.net/browse/ONYX-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/graphql/resolvers/create_submission_resolver.rb
+++ b/app/graphql/resolvers/create_submission_resolver.rb
@@ -6,7 +6,8 @@ class CreateSubmissionResolver < BaseResolver
       SubmissionService.create_submission(
         @arguments,
         @context[:current_user],
-        is_convection: false
+        is_convection: false,
+        access_token: @context[:jwt_token]
       )
 
     {consignment_submission: submission}
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -14,7 +14,8 @@ class SubmissionService
       submission_params,
       gravity_user_id,
       is_convection: true,
-      current_user: nil
+      current_user: nil,
+      access_token: nil
     )
       if submission_params[
         :edition_size_formatted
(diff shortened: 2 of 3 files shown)


MounirDhahri (commented):
Can artworks be linked to many draft submissions? if so which one should we return

MounirDhahri (approved):
looks good to me then - feel free to merge this. Do you mind if you also test this out in app since you have the full context here

PR #1507, opened 2024-08-28
fix: allow publishing re-submitted submissions

nickskalkin:
It was discovered during QA session ([link](https://www.notion.so/artsy/Convection-Resubmitted-artwork-cannot-be-listed-b85049f80d5c423f9a2a886c70f8b4d9?pvs=4))

I think I removed "publish" action by mistake. @dblandin pointed out [here](https://github.com/artsy/convection/pull/1497#discussion_r1691478322) that preferred set of actions should be publish / hold and close, but instead of removing "reject" I removed "publish"


dariakoko (approved):
Thank you!

PR #1515, opened 2024-09-05
feat: display additional artwork attributes from MP

dblandin:
Some of the data we persist within the submission flow lives in Gravity, associated with the collector's My Collection artwork.

This commit queries for that data and renders it within a new uneditable section on the submission admin page.

Screenshot 2024-09-05 at 10 12 04


nickskalkin (approved):
👍

PR #1519, opened 2024-09-06
feat: link admin submission collector name to Forque user page

dblandin:
There's an existing link to the Forque user page under Created At but it hasn't been intuitive enough for colleagues to discover it easily.

This commit adds a new link under Collector Info and additionally updates the Created At link to include the collector name.

## Screenshots

Screenshot 2024-09-06 at 12 38 29
Screenshot 2024-09-06 at 12 41 49


nickskalkin (approved):

app/views/admin/submissions/_created_by.html.erb
```diff
@@ -6,7 +6,7 @@
     <% if @submission.admin %>
       <%= mail_to @submission.admin.email, @submission.admin.email, class: "smaller-sidebar-link" %>
     <% elsif @submission.user&.gravity_user_id.present? %>
-      <%= link_to @submission.user&.gravity_user_id, user_management_url(@submission.user&.gravity_user_id), class: "smaller-sidebar-link" %>
+      <%= link_to @submission.name.presence || @submission.user&.gravity_user_id, user_management_url(@submission.user&.gravity_user_id), class: "smaller-sidebar-link" %>
```
nickskalkin, line 9: just noticed small difference, in another file you do `submission.name` without `presence`

dblandin: Ahh, thanks — I'll update both to use `#presence` as that should catch any strings with only whitespace.

PR #1520, opened 2024-09-06
feat: hyperlink admin submission artist name to Artsy page

dblandin:
Colleagues describe needing to copy/paste the artist name into Artsy and/or Google in order to find artist details. Let's make it easy to at least navigate to the known Artsy page.

https://github.com/user-attachments/assets/af3ea3f2-d64d-4294-91b1-0c1e2ddcf7f7

### DIFF
--- a/app/views/admin/submissions/_submission_details.html.erb
+++ b/app/views/admin/submissions/_submission_details.html.erb
@@ -3,7 +3,7 @@
     Artist/Designer
   </div>
   <div class='overview-item-value'>
-    <%= @artist&.[](:name)&.upcase %>
+    <%= link_to @artist&.[](:name)&.upcase, artist_url(@artist&.[](:slug)), class: "make-primary-link"  %>
   </div>
 </div>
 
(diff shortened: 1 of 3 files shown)


app/views/admin/submissions/_submission_details.html.erb
```diff
@@ -3,7 +3,7 @@
     Artist/Designer
   </div>
   <div class='overview-item-value'>
-    <%= @artist&.[](:name)&.upcase %>
+    <a class="make-primary-link" href="<%= artist_url(@artist&.[](:slug)) %>"><%= @artist&.[](:name)&.upcase %></a>
```
nickskalkin, line 6: **nitpick**: is there a reason to not use [link_to](https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to) from rails?

nickskalkin (approved):

app/views/admin/submissions/_submission_details.html.erb
dblandin: Ahh, no. Good call. Will make that change 👍🏼

PR #1512, opened 2024-08-29
feat: add submission approved email to assigned admin

dblandin:
Send an email to the assigned admin representative once a submission has been approved.

PR #1533, opened 2024-11-08
feat: Send submission resubmitted email to assigned admin

olerichter00:
Addresses [ONYX-1192]

* Related PR that adds an admin email for approved submissions: https://github.com/artsy/convection/pull/1512
* Feature Flag: https://unleash.artsy.net/projects/default/features/onyx-admin-submission-resubmitted-email

## Description

Send an email to the assigned admin representative once a submission has been resubmitted by the user (tier 2 sell flow finished).

[ONYX-1192]: https://artsyproduct.atlassian.net/browse/ONYX-1192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -192,7 +192,7 @@ class SubmissionService
         assigned_admin = AdminUser.find_by(gravity_user_id: submission.assigned_to)
 
         if assigned_admin && Convection.unleash.enabled?(
-          "onyx-admin-submission-approved-email",
+          "onyx-admin-submission-resubmitted-email",
           Unleash::Context.new(user_id: assigned_admin.gravity_user_id.to_s)
         )
           delay.deliver_admin_submission_resubmitted_notification(submission.id)
@@ -336,9 +336,8 @@ class SubmissionService
 
     def deliver_admin_submission_resubmitted_notification(submission_id)
       submission = Submission.find(submission_id)
-      artist = Gravity.client.artist(id: submission.artist_id)._get
 
-      AdminMailer.submission_resubmitted(submission: submission, artist: artist)
+      AdminMailer.submission_resubmitted(submission: submission)
         .deliver_now
     end
 
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -122,6 +122,8 @@ class SubmissionService
       case submission.state
       when "submitted"
         submit!(submission)
+      when "resubmitted"
+        resubmit!(submission)
       when "approved"
         approve!(submission, current_user, is_convection)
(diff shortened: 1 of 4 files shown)


app/services/submission_service.rb
```diff
@@ -185,6 +187,19 @@ def submit!(submission)
         .notify_user(submission.id)
     end
 
+    def resubmit!(submission)
+      if submission.assigned_to
+        assigned_admin = AdminUser.find_by(gravity_user_id: submission.assigned_to)
+
+        if assigned_admin && Convection.unleash.enabled?(
+          "onyx-admin-submission-approved-email",
+          Unleash::Context.new(user_id: assigned_admin.gravity_user_id.to_s)
```
olerichter00, line 196: I haven't created a new feature flag for the new email yet and wonder whether we need one. In my opinion, a feature flag shouldn't be necessary because this email is only sent to admins, and we can easily roll back or fix the code.

@dblandin, you have created a feature flag for the "submission approved" admin email in https://github.com/artsy/convection/pull/1512. Do you think it would be a good idea to create a feature flag for the "Resubmitted" email as well?

nickskalkin: yeah, I think we can easily rollback if anything. Do you know if admins know about upcoming change?

app/services/submission_service.rb
```diff
@@ -319,6 +334,14 @@ def deliver_approval_notification(submission_id)
         .deliver_now
     end
 
+    def deliver_admin_submission_resubmitted_notification(submission_id)
+      submission = Submission.find(submission_id)
+      artist = Gravity.client.artist(id: submission.artist_id)._get
+
+      AdminMailer.submission_resubmitted(submission: submission, artist: artist)
```
nickskalkin, line 341: Is there a need for `artist` here? I've noticed that `AdminMailer.submission_resubmitted` only has `submission` argument`

app/services/submission_service.rb
olerichter00: An Unleash feature flag that can be enabled per user ID might be good to test the new email before sending it to admins. So I think I will create one (and remove it when the feature is released).

Thanks for the reminder to inform about the new email before releasing it 🙇 I haven't done it yet.

nickskalkin (approved):

olerichter00:
QA'd the new email in this is how it looks like ✅ 

Bildschirmfoto 2024-11-28 um 16 49 50

PR #1534, opened 2024-11-11
chore: switch to artsy-event_publisher for publishing rabbitmq events

joeyAghion:
This PR migrates one more project from `artsy-eventservice` to `artsy-event_publisher`, as part of [this broader migration](https://www.notion.so/artsy/Artsy-EventService-EventPublisher-11acab0764a080f5a8b7d9537d97da55) to make event-publishing simpler and more testable.

Convection publishes just 2 events, so I switched them both to the equivalent calls in the new library and dropped the old library and its supporting event classes. No changes should be necessary to configuration or subscribers to these events.


nickskalkin (approved):
thanks!

PR #1537, opened 2024-11-22
chore: Upgrade to sentry-ruby, rails, sidekiq sdks, removes sentry-raven

jpotts244:
Attempt to get us off of deprecated `sentry-raven` gem, installs `sentry-ruby`, `sentry-rails`, `sentry-sidekiq`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   helper Watt::Engine.helpers
 
   before_action :set_current_user
-  before_action :set_raven_context
+  before_action :set_sentry_context
 
   NotAuthorized = Class.new(StandardError)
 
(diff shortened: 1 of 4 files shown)


config/initializers/sentry.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-Raven.configure do |config|
+Sentry.init do |config|
   config.dsn = Convection.config.sentry_dsn if Convection.config.sentry_dsn
-  config.processors -= [Raven::Processor::PostData]
```
jpotts244, line 5: If Im reading the migration docs right, I think this is OK to remove as these processors are no longer used in the sdk and there's default data scrubbing behaviors

https://docs.sentry.io/platforms/ruby/migration/#removed-processors

app/controllers/application_controller.rb
```diff
@@ -26,7 +26,8 @@ def set_current_user
     @current_user = ArtsyAdminAuth.decode_user(session[:access_token])
   end
 
-  def set_raven_context
-    Raven.user_context(id: @current_user)
+  def set_sentry_context
+    # Heads up! @current_user is the user's id as a string not an object
+    Sentry.set_user(user_id: @current_user)
   end
```
jpotts244, line 32: Didnt want to mess with `@current_user` at this time - it is used everywhere 
and its technically a STRING and not an object - bit of a gotcha

jonallured (approved):
LGTM!! 🎉

PR #1544, opened 2024-11-29
chore: Remove Submission mailer feature flags

olerichter00:
Addresses [ONYX-1388]

## Description

This removes all Submission mailer feature flags. The new email templates have already been tested and released. In case we want to disable the mails, we can just roll back the PRs.

[ONYX-1388]: https://artsyproduct.atlassian.net/browse/ONYX-1388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/submission_service.rb
+++ b/app/services/submission_service.rb
@@ -208,7 +208,7 @@ class SubmissionService
         Submission::APPROVED
       )
 
-        delay.deliver_approval_notification(submission.id)
+      delay.deliver_approval_notification(submission.id)
 
       if submission.assigned_to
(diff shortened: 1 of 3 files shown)


dblandin (commented):
I think we still need a guard somewhere to avoid sending an email in the case where no admin is currently assigned to the submission.

Either here in the SubmissionService or a no-op within AdminMailer [here](https://github.com/artsy/convection/blob/f8c9b567fb7ca06e2438ea3e7b0f5f8f23a0f03d/app/mailers/admin_mailer.rb#L33).

PR #1548, opened 2024-12-04
chore: Upgrade to Ruby 3.1

jonallured:
Pursuant to our pattern of upgrading Ruby and then Rails this PR does the Ruby 3.1 upgrade in preparation for Rails 7.0. I had to upgrade Standard Ruby and also add a gem called "matrix" to the Gemfile which wasn't something I had bumped into before. 🤷 

/cc @artsy/amber-devs


jpotts244:
Hi!!! I think we might need to bump selenium-webdrivers version here too

This is a very helpful PR! https://github.com/artsy/volt/pull/7459/files

config/application.rb
```diff
@@ -4,6 +4,7 @@
 
 require "rails/all"
 require_relative "../lib/jwt_middleware"
+require "ostruct"
```
jpotts244, line 8: lol

jonallured: Ok you got me - that's probably a symptom of something else but it fixed a bunch of these errors:

```
NameError:
  uninitialized constant GraphQL::Compatibility::ExecutionSpecification::SpecificationSchema::OpenStruct
```

jpotts244: Im laughing at ruby/removing this from the core library - its everywhere

jonallured:
Closing stale PRs 🧽

lidimayra:
@jonallured thank you for caring on doing this 💟

PR #1550, opened 2024-12-05
feat: load secrets from Vault

artsyjian:
[PHIRE-807]

Load secrets from Vault, per [playbook](https://www.notion.so/artsy/Migration-of-Artsy-Projects-Secrets-07e5ce9e1ed044dba8d4b7c3f0ae1ccf#13fcab0764a080e2be4df5529deb7838).

[PHIRE-807]: https://artsyproduct.atlassian.net/browse/PHIRE-807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -163,11 +163,11 @@ spec:
             periodSeconds: 30
             timeoutSeconds: 3
           volumeMounts:
-            - mountPath: /tmp
-              name: tmp
+          - mountPath: /tmp
+            name: tmp
       volumes:
-        - name: tmp
-          emptyDir: {}
+      - name: tmp
+        emptyDir: {}
       dnsPolicy: ClusterFirst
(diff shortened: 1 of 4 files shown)


hokusai/production.yml
```diff
@@ -298,9 +396,16 @@ spec:
             - partners:daily_digest
```
narikazu, line 396: question (non-blocking): Don't we need to replace `command` with `args` for the rake task?

narikazu (approved):
Left one question but it looks good to me. 
LGTM

hokusai/production.yml
artsyjian: good catch. updated.

PR #1554, opened 2024-12-16
fix: load secrets in data-sync cron containers

nickskalkin:
I've noticed [this error](https://artsy.slack.com/archives/C0HP61PUJ/p1734325328089029). Seems like secrets are not getting loaded in db sync containers. The fix is the same as https://github.com/artsy/positron/pull/3165

I haven't tested changes, just did a copy-paste from positron. Also I don't understand why it's required to load secrets explicitly only fro some containers, but fine to omit for others, could someone explain?

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -589,6 +589,10 @@ spec:
             - name: secrets
               mountPath: /secrets
               readOnly: true
+            - name: scripts
+              mountPath: /usr/local/bin/load_secrets_and_run.sh
+              subPath: load_secrets_and_run.sh
+            command: ["/usr/local/bin/load_secrets_and_run.sh"]
             args:
             - sh
             - ./export-db.sh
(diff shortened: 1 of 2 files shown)


artsyjian (approved):
looks great to me!

amonkhouse:
@artsyjian how come some projects need this?

artsyjian:
> @artsyjian how come some projects need this?

yeah, just added a [section](https://www.notion.so/artsy/Migration-of-Artsy-Projects-Secrets-07e5ce9e1ed044dba8d4b7c3f0ae1ccf?pvs=4#15ecab0764a080269f87f9b33c54eb1a) to the playbook for this.

amonkhouse:
> > @artsyjian how come some projects need this?
> 
> yeah, just added a [section](https://www.notion.so/artsy/Migration-of-Artsy-Projects-Secrets-07e5ce9e1ed044dba8d4b7c3f0ae1ccf?pvs=4#15ecab0764a080269f87f9b33c54eb1a) to the playbook for this.

oh so the `src/load/load.py` just writes `export KEY=VALUE`s to a file, but we haven't actually been setting them in the env until this change. how are other projects using the secrets file that comes from `src/load/load.py`, or are they not?

amonkhouse:
@nickskalkin I think I understand now. Looks like in [other repositories](https://github.com/artsy/candela/pull/327/files) this `load_secrets_and_run.sh` script has been added to the entrypoint in the Docker files, but this is using the data sync container which is defined in a separate repo, so we can't add that here. Although I'm not 100% sure why we can't add it the [pg-data-sync Dockerfile](https://github.com/artsy/docker-images/blob/55d4af65c47b499d2716081f14f97de38330e4fc/pg-data-sync/Dockerfile), could someone explain?

PR #1558, opened 2025-01-09
chore: rescue_from NotAuthorized error in admin controllers

dblandin:
Rescues from any unhandled NotAuthorized exception by rendering a standard 403: Forbidden HTTP response. Our RestController rescues from this exception in a similar way but returns a JSON payload ([link][rest-controller]).

This basic forbidden response is also used by the artsy-auth gem ([link][artsy-auth]).

[rest-controller]: https://github.com/artsy/convection/blob/a2056fac53643ba8f91871d776b4f005a627edab/app/controllers/api/rest_controller.rb#L14-L16
[artsy-auth]: https://github.com/artsy/artsy-auth/blob/53ce676122243748ef2104e3c7fc083698a4f5a5/lib/artsy-auth/authenticated.rb#L13

## Screenshots

| Before | After |
|--------|--------|
| Screenshot 2025-01-09 at 17 15 26 | Screenshot 2025-01-09 at 17 09 03 |

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,10 @@ class ApplicationController < ActionController::Base
 
   expose(:size) { (params[:size] || 100).to_i }
 
+  rescue_from ApplicationController::NotAuthorized do
+    head(:forbidden)
+  end
+
   # override application to decode token and allow only users with `admin` role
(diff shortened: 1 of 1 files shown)


olerichter00 (approved):
Very good idea 👍

PR #1568, opened 2025-02-11
chore: disable some emails to admin & user

anandaroop:
- https://artsyproduct.atlassian.net/browse/ONYX-1541
- https://artsyproduct.atlassian.net/browse/ONYX-1542

For now, disables:
- emails that go the admin email address (sell@)
- submission confirmations that are delivered to the user (rejection emails are untouched)

Instead it logs one of the following messages (for easy inspection of Papertrail, in case there's some code path that still leads to this outcome):

```
[Consignments suspended] Declining to deliver admin email for Submission 42
```

```
[Consignments suspended] Declining to deliver user email for Submission 42
```

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -4,26 +4,28 @@ class UserMailer < ApplicationMailer
   helper :url, :submissions, :offers
 
   def submission_receipt(submission:, artist:)
-    @submission = submission
-    @artist = artist
+    Rails.logger.warn "[Consignments suspended] Declining to deliver user email for Submission #{submission&.id}"
 
-    @utm_params =
-      utm_params(
-        source: "sendgrid",
-        campaign: "sell",
-        term: "cx",
-        content: "received"
-      )
-
-    smtpapi category: %w[submission_receipt],
-            unique_args: {
-              submission_id: submission.id
-            }
-    mail(
-      to: submission.email,
-      subject: "Thank you for submitting your artwork to Artsy",
-      bcc: Convection.config.bcc_email_address
-    )
+    # @submission = submission
+    # @artist = artist
+    #
+    # @utm_params =
+    #   utm_params(
+    #     source: "sendgrid",
+    #     campaign: "sell",
+    #     term: "cx",
+    #     content: "received"
+    #   )
+    #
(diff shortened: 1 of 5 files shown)


spec/requests/api/submissions/update_spec.rb
```diff
@@ -125,17 +125,19 @@
...
+        # admin_copy = "We have received the following submission from: Jon"
+        # expect(admin_email.html_part.body.to_s).to include(admin_copy)
+        # expect(admin_email.text_part.body.to_s).to include(admin_copy)
 
         user_email = emails.detect { |e| e.to.include?("michael@bluth.com") }
-        user_copy =
-          "to evaluate whether we currently have a suitable market for it"
-        expect(user_email.html_part.body.to_s).to include(user_copy)
-        expect(user_email.text_part.body.to_s).to include(user_copy)
+        pp ["ROOP", user_email]
```
dblandin, line 136: Looks like a stray print line here ☝🏼

app/mailers/user_mailer.rb
```diff
@@ -4,26 +4,28 @@ class UserMailer < ApplicationMailer
...
+    #           submission_id: submission.id
+    #         }
+    # mail(
+    #   to: submission.email,
+    #   subject: "Thank you for submitting your artwork to Artsy",
+    #   bcc: Convection.config.bcc_email_address
+    # )
   end
 
   def first_upload_reminder(submission:)
```
dblandin, line 31: **suggestion**: Thoughts on including these first and second reminder emails in this initial disable sweep? We might then also want to include the name of the email in the console warn copy.

dblandin (approved):
LGTM as a next step 👍🏼 

We can do a more thorough pass, with deleted lines and templates, as a follow-up.

app/mailers/user_mailer.rb
anandaroop: Good call, ticketed here for follow-up: https://artsyproduct.atlassian.net/browse/ONYX-1546

spec/requests/api/submissions/update_spec.rb
anandaroop: Weird, I don't see that locally, maybe I got too clever with my rebases 😅 I'll sort that out and merge…

PR #1578, opened 2025-05-09
chore: remove data sync crons

amonkhouse:
These have been migrated to the new staging data sync DAG in [motion](https://github.com/artsy/motion/blob/main/dags/staging_data_sync/pipeline.py#L112-L151).


yhsuanwu (approved):
🚀

